PR #3, opened 2019-09-25
add resourcelock, lease for 4.2->4.3 upgrades

sallyom:
see this also: https://github.com/openshift/openshift-controller-manager/pull/28
@stlaz


deads2k:
@soltysh you probably want to let me be an approver.

PR #5, opened 2019-09-30
remove references to config related to legacy controllers

sanchezl:
The `OpenShiftControllerManagerConfig` aggregated the configuration of various controllers. Only the resource quota and namespace security allocation controllers remain. Remove all references to configuration not used by the remaining controllers.

The remaining configuration is:
```yaml
# common
kubeClientConfig:
servingInfo:
leaderElection:
controllers:
# controller specific
resourceQuota:
securityAllocator:
```


deads2k:
/lgtm

@sanchezl make us a new configuration type if we have options for it.

PR #6, opened 2019-10-08
make --config optional

sanchezl:
1. We've been using an empty config file.
2. The internal defaults are sufficient.
3. We offer no ability to configure anyway.

### DIFF
--- a/pkg/cmd/cluster-policy-controller/cmd.go
+++ b/pkg/cmd/cluster-policy-controller/cmd.go
@@ -1,7 +1,6 @@
 package cluster_policy_controller
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -67,19 +63,10 @@ func NewClusterPolicyControllerCommand(name string, out, errout io.Writer) *cobr
 	// This command only supports reading from config
 	flags.StringVar(&options.ConfigFilePath, "config", options.ConfigFilePath, "Location of the master configuration file to run from.")
 	cmd.MarkFlagFilename("config", "yaml", "yml")
-	cmd.MarkFlagRequired("config")
 
 	return cmd
 }
 
-func (o *ClusterPolicyController) Validate() error {
-	if len(o.ConfigFilePath) == 0 {
```
deads2k: leave the method, delete the content.

deads2k:
/approve

close enough and we need to get unstuck

PR #9, opened 2019-10-15
allow --kubeconfig files

deads2k:


sallyom:
Also needs HTTPServingInfo.ServingInfo.CertInfo but this gets a little further

PR #10, opened 2019-10-15
Switch to standard controller startup

deads2k:


deads2k:
this may actually make it harder to do the delta.

/hold

deads2k:
Yeah, this won't happen until after we get it running and someone else can do it.

/close

PR #11, opened 2019-10-15
Add config params

sallyom:
testing w/ this now
@deads2k this is being tested [here](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/287)


sallyom:
this won't be required, this is much better: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/296

PR #16, opened 2020-01-20
Switch to go.mod and bump to k8s 1.17.1

soltysh:
Picks some of the changes from https://github.com/openshift/cluster-policy-controller/pull/10 but I started with go.mod and bump first. More will follow up to land something along the lines like #10.
/assign @deads2k


go.mod
```diff
@@ -0,0 +1,60 @@
...
+	k8s.io/apimachinery v0.17.1
+	k8s.io/apiserver v0.17.1
+	k8s.io/client-go v0.17.1
+	k8s.io/component-base v0.17.1
+	k8s.io/klog v1.0.0
+	k8s.io/kubectl v0.0.0
+	k8s.io/kubernetes v1.17.1
+)
+
+replace (
```
deads2k: later try to remove replaces.  My guess is that k/k is requiring several of these replaces.  If so, just comment to that effect.

soltysh: I already tried, the ones required are the top 4, although I might try getting rid of glog now for sure, probably ginkgo too the other two has to stay. The k8s needs to be there since go.mod can't parse them out from k/k since it points to staging :disappointed:

soltysh:
Fixed deps, let's wait for the tests now.

PR #18, opened 2020-01-24
Bug 1794366: Fix GOFLAGS in dockerfile

mfojtik:
This unblocks CI job that is failing because cluster-policy-controller binary is not present in payload image.

This should be reverted and handled properly later.


soltysh:
/hold
for discussion

PR #20, opened 2020-03-11
[release-4.4] Bug 1812594: Live reload kubeconfig client certs

tnozicka:
cluster-policy-controller is coupled into kube-controller-manager pod which is wired for live reload of client certs so it must handle reload the client certs without a rollout or it is gonna get stuck with old and invalid client certs.

backporting https://github.com/openshift/cluster-policy-controller/pull/19

/cc @deads2k @soltysh


tnozicka:
I need to re-open it to the 4.4 branch as that can't be changed :(

PR #22, opened 2020-04-10
Bug 1820687: use patch when updating namespace

soltysh:
/assign @deads2k 

Requires https://github.com/openshift/cluster-kube-controller-manager-operator/pull/394
/hold

### DIFF
--- a/pkg/security/controller/namespace_scc_allocation_controller.go
+++ b/pkg/security/controller/namespace_scc_allocation_controller.go
@@ -6,19 +6,23 @@ import (
 	"reflect"
 	"time"
 
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog"
-
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -178,7 +182,21 @@ func (c *NamespaceSCCAllocationController) allocate(ns *corev1.Namespace) error
 		}
 	}
 
-	_, err = c.namespaceClient.Update(nsCopy)
+	encoder := legacyscheme.Codecs.LegacyCodec(legacyscheme.Scheme.PrioritizedVersionsAllGroups()...)
```
deads2k, line 185: install the core types into a scheme to do this.  You want to snip k/k links, not create them.

deads2k:
just the note about the scheme, then lgtm

/cherrypick release-4.4

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -54,6 +60,13 @@ func NewNamespaceSCCAllocationController(
 	requiredUIDRange *uid.Range,
 	mcs MCSAllocationFunc,
 ) *NamespaceSCCAllocationController {
+
+	scheme := runtime.NewScheme()
```
deads2k, line 64: you can do all this in an init block and have the encoder as a package var

PR #26, opened 2020-04-29
[release-4.3] Bug 1829328: use patch when updating namespace

soltysh:
4.3 pick of #22
/assign @deads2k


soltysh:
/hold
for https://github.com/openshift/openshift-apiserver/pull/101

sdodson:
The bug is only medium severity but this fixes a test case so accepting it.

PR #28, opened 2020-06-05
Bug 1808588: add UID deallocation logic

soltysh:
/assign @deads2k @smarterclayton 
/hold
since this is delicate and requires careful testing and review

### DIFF
--- a/pkg/security/controller/namespace_scc_allocation_controller.go
+++ b/pkg/security/controller/namespace_scc_allocation_controller.go
@@ -34,8 +34,9 @@ import (
 )
 
 const (
-	controllerName = "namespace-security-allocation-controller"
-	rangeName      = "scc-uid"
+	controllerName    = "namespace-security-allocation-controller"
+	rangeName         = "scc-uid"
+	periodicRepairKey = "__internal/periodicRepair"
 )
 
 // NamespaceSCCAllocationController allocates uids/labels for namespaces
@@ -109,6 +110,8 @@ func (c *NamespaceSCCAllocationController) Run(stopCh <-chan struct{}) {
 		klog.Fatal(err)
 	}
 	klog.V(1).Infof("Repair complete")
+
+	c.AddNextPeriodicRepair()
 
 	go c.worker()
 	<-stopCh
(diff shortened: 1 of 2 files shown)


soltysh:
I've tried this in my local cluster and works fine, as soon as this merges I'll add an e2e to origin to ensure it's working as expected.

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -216,6 +220,35 @@ func (c *NamespaceSCCAllocationController) allocate(ns *corev1.Namespace) error
 	return nil
 }
 
+func (c *NamespaceSCCAllocationController) deallocate(ns *corev1.Namespace) error {
```
smarterclayton, line 223: This is insufficient.  If this controller is stopped while a namespace is being deleted, then no one will ever fix it and eventually you will leak enough indices that you run out of space.

You MUST rebuild the state of the world periodically and do a read, check, replace.  If you race with another instance of your controller the update method will fail and you just reenqueue the repair.

Please use repair correctly - repair is safe if Get -> Update does not return a 409.

smarterclayton:
/hold

This is not correct in the presence of the controller being down and will leak.

pkg/security/controller/namespace_scc_allocation_controller.go
soltysh: > This is insufficient. If this controller is stopped while a namespace is being deleted, then no one will ever fix it and eventually you will leak enough indices that you run out of space.

If it's stopped it will call `Repair` upon restart, in which case it should fix the current state of the world. I'm not removing the `Repair` I'm just adding ad-hoc deallocation. 


> You MUST rebuild the state of the world periodically and do a read, check, replace. If you race with another instance of your controller the update method will fail and you just reenqueue the repair.

Yeah, that's fair. I didn't want to call `Repair` on every deletion since that might be taking too long, but it's definitely worthy to call this periodically.

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -239,7 +239,13 @@ func (c *NamespaceSCCAllocationController) AddNextPeriodicRepair() {
 	// For this controller to work correctly we need to ensure a periodic repair
 	// of all the range allocations. For that we are adding an artificial key
 	// which will trigger that every 8 hours.
-	delta := time.Duration(8 * time.Hour)
+
+	// ****************************************************************************
```
soltysh, line 243: This is explicitly in a separate commit and I have https://bugzilla.redhat.com/show_bug.cgi?id=1856687 to revert that change before shipping 4.6.

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -226,6 +235,21 @@ func allocateNextContiguousBit(allocated *big.Int, max int) (int, bool) {
...
 }
 
+func (c *NamespaceSCCAllocationController) AddNextPeriodicRepair() {
+	// For this controller to work correctly we need to ensure a periodic repair
+	// of all the range allocations. For that we are adding an artificial key
+	// which will trigger that every 8 hours.
+
+	// ****************************************************************************
+	// For the development cycle, make the periodic repair faster (every 5 minutes)
+	delta := time.Duration(5 * time.Minute)
```
smarterclayton, line 245: Were we not going to add some sort of global private config that overrides this?

smarterclayton:
Looks fine.

pkg/security/controller/namespace_scc_allocation_controller.go
soltysh: That was my initial idea, but I dropped it. This mechanism proved to work just fine in kube-controller-manager-operator.

soltysh:
I spoke with Clayton about it on slack, he's ok with the current implmentation. 
/hold cancel

PR #30, opened 2020-06-22
Wait for authorization update in kube apiserver

mfojtik:


mfojtik:
/hold

this won't solve the cae when the apiserver is restarted during leader election wait

deads2k:
This doesn't appear to be necessary.

/close

PR #39, opened 2020-09-28
Bug 1883458: switch to CRD for rangeallocation

deads2k:
based on https://github.com/openshift/api/pull/751

We see numerous SCC pod failures across many namespaces: https://github.com/openshift/origin/pull/25544 .

In the cluster-policy-controller, we see failures about an inability to manage rangeallocations. This is because it is served as an aggregated API from openshift-apiserver.  We moved SCC to a CRD, but forgot it's dependencies.  This PR creates a new RangeAllocation CRD that is schematically identical to the existing one, but now in the `security.internal.openshift.io` group.  This does two things
 1. indicates that it is internal
 2. avoids requiring kube-apiserver patches to expose the CRD.  Co-locating the in same group would require this.

In 4.6, we can change to the CRD without intermediate migration.  This is possible because the cluster-policy-controller runs using a leader elected lease and its first act is to reconcile the rangeallocations with the namespaces that currently exist.  See https://github.com/openshift/cluster-policy-controller/blob/eda513e6d0e645f44deb1c9b4407e46db7cec23a/pkg/security/controller/namespace_scc_allocation_controller.go#L92-L95

we
1. merge this PR
2. vendor into cluster-config-operator
3. add to client-go
4. switch client in cluster-policy-controller

The change is relatively low risk.


pkg/security/controller/repair_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"math/big"
 	"testing"
 
+	securityinternalv1 "github.com/openshift/api/securityinternal/v1"
```
sttts, line 7: order

sttts:
A nit.

Plus needs real bump.

/lgtm
/hold

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -29,7 +29,8 @@ import (
 	coreapi "k8s.io/kubernetes/pkg/apis/core"
 
 	securityv1 "github.com/openshift/api/security/v1"
-	securityv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
+	securityinternalv1 "github.com/openshift/api/securityinternal/v1"
+	securityv1client "github.com/openshift/client-go/securityinternal/clientset/versioned/typed/securityinternal/v1"
```
soltysh, line 33: nit: since you're using both security and securityinternal, I'd prefer this have securityinternalv1client - long but clear it's the internal

deads2k:
made the bump real.  relabelling.

PR #40, opened 2020-10-03
add event when SCC ranges are created for namespace

deads2k:
We still events when SCCs cannot be used.  While it appears these are quickly resolved, this event will provide solid evidence.


soltysh (approved):
/lgtm
/retest

PR #42, opened 2020-10-15
Bug 1890530: release lock on graceful shutdown

deads2k:
This will improve responsiveness during bootstrapping and on upgrade.


soltysh (approved):
/lgtm

PR #44, opened 2020-10-28
runs OpenShift specific ResourceQuota monitors

p0lyn0mial:
for backward compatability we need to maintain "openshift.io/imagestreams" alias for "count/imagestreams.image.openshift.io"
to dynamically update quota for imagestreams resources.

### DIFF
--- a/pkg/cmd/controller/quota.go
+++ b/pkg/cmd/controller/quota.go
@@ -7,8 +7,12 @@ import (
 	"github.com/openshift/cluster-policy-controller/pkg/quota/clusterquotareconciliation"
 	image "github.com/openshift/cluster-policy-controller/pkg/quota/quotaimageexternal"
 	"github.com/openshift/library-go/pkg/quota/clusterquotamapping"
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/controller/quota.go
```diff
@@ -46,6 +53,34 @@ func RunResourceQuotaManager(ctx *ControllerContext) (bool, error) {
...
+				return nil, err
+			}
+			for i := range resourceList.APIResources {
+				gr := schema.GroupResource{
+					Group:    gv.Group,
+					Resource: resourceList.APIResources[i].Name,
+				}
+				if evaluator := registry.Get(gr); evaluator != nil {
+					interestingResources = append(interestingResources, resourceList)
+					break
```
p0lyn0mial, line 76: maybe we should remove it, what if more resources in the group apply?

soltysh (approved):
/lgtm
/approve

soltysh (approved):
/lgtm
/hold cancel

PR #61, opened 2021-06-10
AUTH-26 CSR approving mechanism

slaskawi:
https://issues.redhat.com/browse/AUTH-26

This Pull Request introduces CSR Approver (see https://github.com/openshift/library-go/pull/1083) into the Cluster Policy Controller.


pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,44 @@
...
+
+	eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), infraClusterCSRApproverControllerServiceAccountName, &v1.ObjectReference{})
+
+	controller := csr.NewCSRApproverController(
+		controllerName,
+		operatorClient,
+		kubeClient.CertificatesV1().CertificateSigningRequests(),
+		ctx.KubernetesInformers.Certificates().V1().CertificateSigningRequests(),
+		csr.NewLabelFilter(labels.Everything()),
+		csr.NewServiceAccountApprover(monitoringServiceAccountNamespace, monitoringServiceAccountName, monitoringCertificateSubject),
```
slaskawi, line 38: @stlaz @s-urbaniak Both the CSR Filter and the Service Account Approver seem to do the same thing (in terms of functionality) - they filter somehow what CSRs are being approved. Perhaps we should create a Service Account Filter for them and squash both things into a single interface?

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	controllerName = "csr-approver-controller"
+	// Hardcoded for the Monitoring Operator:
+	// https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27
+	monitoringServiceAccountNamespace = "openshift-monitoring"
+	monitoringServiceAccountName = "cluster-monitoring-operator"
+	monitoringCertificateSubject = "/CN=monitoring"
```
slaskawi, line 18: @s-urbaniak Please check those 3 items - are they correct?

pkg/cmd/controller/csr.go
stlaz: They each serve a different purpose, I would be against such a unification

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), infraClusterCSRApproverControllerServiceAccountName, &v1.ObjectReference{})
+
+	controller := csr.NewCSRApproverController(
+		controllerName,
+		operatorClient,
+		kubeClient.CertificatesV1().CertificateSigningRequests(),
+		ctx.KubernetesInformers.Certificates().V1().CertificateSigningRequests(),
+		csr.NewLabelFilter(labels.Everything()),
```
stlaz, line 37: I hope this is a subject to change :slightly_smiling_face:

pkg/cmd/controller/csr.go
stlaz: the subject seems wrong, I would expect it to match the name of the serviceaccount pulling the metrics

s-urbaniak: Yes, the certificate requestor is cluster-monitoring-operator, the target subject is Prometheus, how about `/CN=prometheus`?

pkg/cmd/controller/csr.go
s-urbaniak: what do we want to set here? :thinking:

pkg/cmd/controller/csr.go
s-urbaniak: agreeing with @stlaz, a service account approver is just a very specific filter validating username, groups, and subject.

slaskawi: Ok, understood. Thanks for the context!

pkg/cmd/controller/csr.go
slaskawi: Exactly! So what shall we put here? @stlaz?

pkg/cmd/controller/csr.go
slaskawi: Done

slaskawi:
Rebased and updated. Ready for the first round of reviews.

pkg/cmd/controller/csr.go
stlaz: configure a specific name/label for the CSR

soltysh:
@atiratree fyi, since you were to bump the deps.

pkg/cmd/controller/csr.go
s-urbaniak: Oh, so an opinioated one for the sake of this use case? of course, that makes sense :+1: 
This would prevent auto-approving other CSRs from the same serviceaccont.
How about `openshift.io/metrics`?

atiratree:
@soltysh we still want to update k8s to v0.22 once it is released, right?

pkg/cmd/controller/csr.go
slaskawi: Ok, updated. Please take another look at it.

slaskawi:
@s-urbaniak @stlaz Ready for the next round of reviews.

pkg/operator/operatorclient.go
```diff
@@ -0,0 +1,36 @@
+package operator
+
+import (
+	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
```
s-urbaniak, line 10: nit: whitespace in imports (really a very small nit)

stlaz:
/hold
there shall be no jira ticket names in commit messages

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
+package controller
+
+import (
+	"context"
+	"github.com/openshift/cluster-policy-controller/pkg/operator"
+	"github.com/openshift/library-go/pkg/operator/csr"
+	"github.com/openshift/library-go/pkg/operator/events"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
```
stlaz, line 10: nit: order this such that it's
```
system libs

k8s libs

openshift libs

local libs
```

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	controllerName = "csr-approver-controller"
+	// Hardcoded for the Monitoring Operator:
+	// https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27
+	monitoringServiceAccountNamespace = "openshift-monitoring"
+	monitoringServiceAccountName      = "cluster-monitoring-operator"
+	monitoringCertificateSubject      = "/CN=prometheus"
```
stlaz, line 19: @s-urbaniak I would've expected `/CN=system:serviceaccount:openshift-monitoring:openshift-state-metrics`, no?

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func RunCSRApproverController(ctx *ControllerContext) (bool, error) {
+	kubeClient, err := ctx.ClientBuilder.Client(infraClusterCSRApproverControllerServiceAccountName)
+	if err != nil {
+		return true, err
+	}
+
+	operatorClient := &operator.OperatorClient{}
```
stlaz, line 28: we'll want to get https://github.com/openshift/library-go/pull/1105 merged first to use it here

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+
+	eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), infraClusterCSRApproverControllerServiceAccountName, &v1.ObjectReference{})
+
+	selector := labels.NewSelector()
+	labelsRequirement, err := labels.NewRequirement("openshift.io/metrics", selection.In, []string{"true"})
```
stlaz, line 33: how about `metrics.openshift.io/csr.subject: openshift-state-metrics`?

pkg/operator/operatorclient.go
```diff
@@ -0,0 +1,36 @@
+package operator
+
+import (
+	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
stlaz, line 5: nit: split system and k8s deps with a newline

pkg/operator/operatorclient.go
```diff
@@ -0,0 +1,36 @@
...
+	return nil, nil, "", fmt.Errorf("not implemented")
+}
+
+func (c OperatorClient) UpdateOperatorSpec(resourceVersion string, spec *operatorv1.OperatorSpec) (*operatorv1.OperatorSpec, string, error) {
+	return nil, "", fmt.Errorf("not implemented")
+}
+
+func (c OperatorClient) UpdateOperatorStatus(resourceVersion string, status *operatorv1.OperatorStatus) (*operatorv1.OperatorStatus, error) {
+	return nil, fmt.Errorf("not implemented")
+}
```
stlaz, line 36: we don't need this if we get https://github.com/openshift/library-go/pull/1105 merged first

stlaz (changes requested):
we'll want to find an approver for https://github.com/openshift/library-go/pull/1105 to include it here

pkg/cmd/controller/csr.go
s-urbaniak: sounds good to me. **does** it have to match the exact service account name? I may be missing something, I assumed this is more or less free-form.

stlaz: I don't think it has to match, although I think I would expect it to. Or at least prepend it with the `system:` prefix

s-urbaniak: I think following the canonical k8s syntax as you suggested sounds good. It is very clear to what it maps to :+1:

slaskawi:
> we'll want to find an approver for openshift/library-go#1105 to include it here

@stlaz I see there are some outstanding comments there, so let me propose to implement this in two-steps:
- We'll leave the code as is
- Once https://github.com/openshift/library-go/pull/1105 gets merged, I'll update the code here.

Would this work for you?

pkg/operator/operatorclient.go
slaskawi: I see @s-urbaniak has commented the other file because there was a blank line :)

@stlaz @s-urbaniak @sttts Do we have a formatter for our IDEs somewhere or at least written guidelines about the code style? Sorting imports manually is a wasteful job. 

If there is no formatter and no written code style guidelines, I suggest to leave this as is and introduce the `goimports` into our Makefiles (of course I volunteer to do that).

pkg/cmd/controller/csr.go
slaskawi: Let's continue this discussion here: https://github.com/openshift/cluster-policy-controller/pull/61#discussion_r654274660

pkg/operator/operatorclient.go
slaskawi: Let's continue this discussion here: https://github.com/openshift/cluster-policy-controller/pull/61#discussion_r654274660

pkg/cmd/controller/csr.go
slaskawi: Done

pkg/cmd/controller/csr.go
slaskawi: I proposed slightly different (more async) approach to this here: https://github.com/openshift/cluster-policy-controller/pull/61#issuecomment-863886456

pkg/cmd/controller/csr.go
slaskawi: Done

pkg/operator/operatorclient.go
slaskawi: I proposed slightly different (more async) approach to this here: https://github.com/openshift/cluster-policy-controller/pull/61#issuecomment-863886456

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+
+	"github.com/openshift/cluster-policy-controller/pkg/operator"
+	"github.com/openshift/library-go/pkg/operator/csr"
+	"github.com/openshift/library-go/pkg/operator/events"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
```
stlaz, line 11: wrong order

pkg/operator/operatorclient.go
stlaz: unfortunately that does not help, as long as the errors are returned here we're not moving forward by merging this PR

stlaz (changes requested):

pkg/operator/operatorclient.go
slaskawi: It turned out that @stlaz uses `goimports`. I'll give it a go.

pkg/operator/operatorclient.go
slaskawi: @stlaz @s-urbaniak So it seems this PR is blocked until https://github.com/openshift/library-go/pull/1105 gets in.

slaskawi: Just in case - I copy-pasted the `NoOpOperatorClient` temporarily into my PR as well.

pkg/cmd/controller/csr.go
slaskawi: Fixed

soltysh:
> @soltysh we still want to update k8s to v0.22 once it is released, right?

Correct, but it should be easier after this :wink:

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const (
+	controllerName = "csr-approver-controller"
+	// Hardcoded for the Monitoring Operator:
+	// https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27
+	monitoringServiceAccountNamespace = "openshift-monitoring"
+	monitoringServiceAccountName      = "cluster-monitoring-operator"
+	monitoringCertificateSubject      = "/CN=system:serviceaccount:openshift-monitoring:openshift-state-metrics"
```
s-urbaniak, line 20: as discussed in https://github.com/openshift/cluster-policy-controller/pull/61#discussion_r649790728 this should rather be `/CN=system:serviceaccount:openshift-monitoring:prometheus`

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+const (
+	controllerName = "csr-approver-controller"
+	// Hardcoded for the Monitoring Operator:
+	// https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27
```
s-urbaniak, line 17: reference the commit id as master will change https://github.com/openshift/cluster-monitoring-operator/blob/e7a2b2e20d19ba7a9271df59e06b1fd32b8bdfe4/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,54 @@
...
+	monitoringServiceAccountNamespace = "openshift-monitoring"
+	monitoringServiceAccountName      = "cluster-monitoring-operator"
+	monitoringCertificateSubject      = "/CN=system:serviceaccount:openshift-monitoring:openshift-state-metrics"
+	monitoringLabelKey                = "metrics.openshift.io/csr.subject"
+	monitoringLabelValue              = "openshift-state-metrics"
```
s-urbaniak, line 22: `prometheus`

slaskawi: Done

pkg/cmd/controller/csr.go
slaskawi: No, I'm not very keen on doing this. If this file gets moved, the ID as SHA-1 will point to a still-existing file. This will give proper pointers where to look next.

pkg/cmd/controller/csr.go
slaskawi: Done

slaskawi:
@stlaz @s-urbaniak Updated and rebased. Ready for the next round.

pkg/cmd/controller/csr.go
stlaz: actually, according to https://github.com/openshift/library-go/blob/master/pkg/authorization/hardcodedauthorizer/metrics.go#L23, we may want to just go with `"system:serviceaccount:openshift-monitoring:prometheus-k8s"`

s-urbaniak: ah yes, i missed the suffix

slaskawi: Done

slaskawi:
Updated and rebased

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+func RunCSRApproverController(ctx *ControllerContext) (bool, error) {
+	kubeClient, err := ctx.ClientBuilder.Client(infraClusterCSRApproverControllerServiceAccountName)
+	if err != nil {
+		return true, err
+	}
+
+	eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), infraClusterCSRApproverControllerServiceAccountName, &v1.ObjectReference{})
+
+	selector := labels.NewSelector()
+	labelsRequirement, err := labels.NewRequirement(monitoringLabelKey, selection.In, []string{monitoringLabelValue})
```
stlaz, line 33: nit: `s/selection.In/selection.Equals/`

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return true, err
+	}
+	selector.Add(*labelsRequirement)
```
stlaz, line 37: I believe this should be `selector = selector.Add(*labelsRequirement)`

https://github.com/enj/cluster-authentication-operator/blob/d303865704cb8d9e0e807e63d2f578ecebb42aa2/vendor/k8s.io/apimachinery/pkg/labels/selector.go#L365-L376

stlaz (changes requested):

pkg/cmd/controller/csr.go
slaskawi: :man_facepalming: Of course! You're right!

pkg/cmd/controller/csr.go
slaskawi: Done

pkg/cmd/controller/csr.go
slaskawi: Done

pkg/cmd/controller/csr.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+const (
+	controllerName = "csr-approver-controller"
+	// Hardcoded for the Monitoring Operator:
+	// https://github.com/openshift/cluster-monitoring-operator/blob/e7a2b2e20d19ba7a9271df59e06b1fd32b8bdfe4/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L25-L27
```
stlaz, line 16: nittiest of nits: let's not put github references in the code, I think the hardcode is apparent here.

btw: I believe the correct reference would be https://github.com/openshift/cluster-monitoring-operator/blob/e7a2b2e20d19ba7a9271df59e06b1fd32b8bdfe4/manifests/0000_50_cluster-monitoring-operator_03-role-binding.yaml#L1-L10

stlaz:
another nit: make the commit message be `CSR approver for monitoring scraper`

pkg/cmd/controller/csr.go
slaskawi: Removed that comment.

soltysh (approved):
/approve
/lgtm

PR #65, opened 2021-07-05
refactor cluster policy controller and  NamespaceSCCAllocationController

atiratree:
- I have refactored the main cluster-policy-controller server logic and RunNamespaceSecurityAllocationController.
- RunResourceQuotaManager controller is imported from kubernetes
- RunClusterQuotaReconciliationController seemed more complex and unsuitable for the factory methods. But I could try to redesign the factory methods if needed 

TODO:
- [x] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/545
- [x] https://github.com/openshift/library-go/pull/1134


pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -1,115 +1,300 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/serviceability"
+
+	// for metrics
+	_ "github.com/openshift/library-go/pkg/controller/metrics"
+
+	clusterpolicyversion "github.com/openshift/cluster-policy-controller/pkg/version"
 )
 
 const RecommendedStartControllerName = "cluster-policy-controller"
```
atiratree, line 36: Is this variable still used for something?

pkg/cmd/cluster-policy-controller/policy_controller.go
```diff
@@ -12,11 +12,14 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/serviceability"
 
 	origincontrollers "github.com/openshift/cluster-policy-controller/pkg/cmd/controller"
 )
 
 func RunClusterPolicyController(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
+	serviceability.InitLogrusFromKlog()
```
atiratree, line 21: I am not sure if we still need this one but I have not found a usage for this.

pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -1,115 +1,300 @@
...
+		DisableServing:        false,
+		DisableLeaderElection: false,
+	}
+}
+
+// NewCommandWithContext returns a new command that a caller must set the Use and Descriptions on.  It wires default log, profiling,
+// leader election and other "normal" behaviors.
+// The context passed will be passed down to controller loops and observers and cancelled on SIGTERM and SIGINT signals.
+// copied from controllercmd.ControllerCommandConfig
+func (c *ClusterPolicyControllerCommandConfig) NewCommandWithContext(ctx context.Context) *cobra.Command {
```
atiratree, line 81: This is copy pasted logic for now, since we need a custom logic in StartController and access to the internal variables.

I will make a PR against library-go so this could be customizable and we could avoid copy paste

soltysh:
you will want to wait for https://github.com/openshift/cluster-policy-controller/pull/61 to land, it's also bumping all deps

atiratree:
> you will want to wait for #61 to land, it's also bumping all deps

will rebase once it merges, but it is missing the bump in `replace`s which is included here

pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -1,115 +1,300 @@
...
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/serviceability"
+
+	// for metrics
+	_ "github.com/openshift/library-go/pkg/controller/metrics"
+
+	clusterpolicyversion "github.com/openshift/cluster-policy-controller/pkg/version"
```
soltysh, line 33: Keep openshift imports in a single group above.

pkg/cmd/cluster-policy-controller/cmd.go
soltysh: I don't think so, if not used feel free to drop it.

pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -1,115 +1,300 @@
...
 
 const RecommendedStartControllerName = "cluster-policy-controller"
 
-type ClusterPolicyController struct {
-	ConfigFilePath string
-	// KubeConfigFile points to a kubeconfig file if you don't want to use the in cluster config
-	KubeConfigFile string
+func NewClusterPolicyControllerCommand(name string) *cobra.Command {
+	cmd := NewClusterPolicyControllerCommandConfig("cluster-policy-controller", clusterpolicyversion.Get(), RunClusterPolicyController).
+		NewCommandWithContext(context.TODO())
```
soltysh, line 40: `context.Background()` ?

pkg/cmd/cluster-policy-controller/policy_controller.go
soltysh: Yeah, looks like this is unused can be dropped.

soltysh (commented):
Leaving some comments, but I'll have  another look once the other merges and this gets rebased. Also please put all the vendor changes into single commit.

pkg/cmd/cluster-policy-controller/policy_controller.go
atiratree: dropped

pkg/cmd/cluster-policy-controller/cmd.go
atiratree: dropped

pkg/cmd/cluster-policy-controller/cmd.go
atiratree: I am little bit confused by ` It is never canceled`: https://golang.org/pkg/context/#Background . Nevertheless, I updated the code.

atiratree:
@soltysh the code was simplified to use only the default config . There are some changes to the operator that need to be merged https://github.com/openshift/cluster-kube-controller-manager-operator/pull/545 . 

What is the order of merging when we need both of these changes at the same time?

atiratree:
rebased and using WithComponentOwnerReference @soltysh

soltysh (approved):
/lgtm
/approve

PR #68, opened 2021-07-23
use correct eventRecorder in CSRApproverController

atiratree:
we can now use an EventRecorder initialized by libary-go after this refactoring https://github.com/openshift/cluster-policy-controller/pull/65


soltysh (approved):
/lgtm
/approve

PR #69, opened 2021-07-23
add namespace SCC allocation metrics

atiratree:
this will add a `openshift_namespace_scc_allocation_total` metric. This metric tracks `Resolved`/`Pending` namespaces that have resolved scc annotations. This would help for example with bugs like these: https://bugzilla.redhat.com/show_bug.cgi?id=1968383

TODO:
- [x] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/554
- [x] https://github.com/openshift/installer/pull/5105
- [ ] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/615

### DIFF
--- a/pkg/cmd/controller/security.go
+++ b/pkg/cmd/controller/security.go
@@ -28,7 +28,6 @@ func RunNamespaceSecurityAllocationController(ctx context.Context, controllerCtx
 	if err != nil {
 		return true, err
 	}
-
 	controller := sccallocation.NewNamespaceSCCAllocationController(
 		controllerCtx.KubernetesInformers.Core().V1().Namespaces(),
 		kubeClient.CoreV1().Namespaces(),
--- a/pkg/security/controller/namespace_scc_allocation_controller.go
+++ b/pkg/security/controller/namespace_scc_allocation_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/openshift/library-go/pkg/operator/events"
 	"math/big"
 	"reflect"
 	"time"
@@ -31,8 +30,10 @@ import (
 	securityinternalv1 "github.com/openshift/api/securityinternal/v1"
 	securityv1client "github.com/openshift/client-go/securityinternal/clientset/versioned/typed/securityinternal/v1"
 	"github.com/openshift/cluster-policy-controller/pkg/security/mcs"
+	sccmetrics "github.com/openshift/cluster-policy-controller/pkg/security/metrics"
 	"github.com/openshift/cluster-policy-controller/pkg/security/uidallocator"
 	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/security/uid"
 )
 
@@ -57,6 +58,7 @@ type NamespaceSCCAllocationController struct {
 	namespaceClient       corev1client.NamespaceInterface
 	rangeAllocationClient securityv1client.RangeAllocationsGetter
 
+	metrics sccmetrics.NamespaceSCCAllocationMetrics
 	encoder runtime.Encoder
 }
 
@@ -67,12 +69,15 @@ func NewNamespaceSCCAllocationController(namespaceInformer corev1informers.Names
 	jsonSerializer := runtimejson.NewSerializer(runtimejson.DefaultMetaFactory, scheme, scheme, false)
 	encoder := codecs.WithoutConversion().EncoderForVersion(jsonSerializer, corev1.SchemeGroupVersion)
 
+	metrics := sccmetrics.NewNamespaceSCCAllocMetrics()
+
 	c := &NamespaceSCCAllocationController{
 		requiredUIDRange:      requiredUIDRange,
 		mcsAllocator:          mcs,
 		namespaceClient:       client,
 		rangeAllocationClient: rangeAllocationClient,
 		nsLister:              namespaceInformer.Lister(),
+		metrics:               metrics,
 		encoder:               encoder,
 	}
 
@@ -98,7 +103,14 @@ func NewNamespaceSCCAllocationController(namespaceInformer corev1informers.Names
 		10*time.Minute,
 	)
 
-	return factory.New().ResyncSchedule(resyncPeriod).WithBareInformers(namespaceInformer.Informer()).WithSyncContext(syncContext).WithSync(c.sync).
+	sccmetrics.Register()
+	klog.V(4).Info("namespace scc allocation metrics registered with prometheus")
+
+	return factory.New().
+		ResyncSchedule(resyncPeriod).
+		WithBareInformers(namespaceInformer.Informer()).
+		WithSyncContext(syncContext).
+		WithSync(c.sync).
 		ToController(controllerName, eventRecorderWithSuffix)
 }
 
@@ -188,6 +200,7 @@ func (c *NamespaceSCCAllocationController) allocate(ctx context.Context, syncCtx
 		return err
 	}
 	c.currentUIDRangeAllocation = actualRangeAllocation
+	c.metrics.UIDRangeUpdated(actualRangeAllocation)
 
 	block, ok := uidRange.BlockAt(uint32(bitIndex))
 	if !ok {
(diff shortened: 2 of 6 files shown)


pkg/security/metrics/prometheus/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+
+	securityv1 "github.com/openshift/api/security/v1"
+)
+
+const (
+	separator                   = "_"
+	namespaceSCCAllocSubsystem  = "openshift_namespace_scc_allocation"
+	namespaceSCCAllocCount      = "total"
+	namespaceSCCAllocCountQuery = namespaceSCCAllocSubsystem + separator + namespaceSCCAllocCount
+)
```
atiratree, line 22: would it also make sense to collect number or percentage of uids allocated?

pkg/security/metrics/prometheus/metrics.go
```diff
@@ -0,0 +1,123 @@
+package prometheus
+
+import (
+	"sync"
+
+	"github.com/blang/semver"
+	"github.com/prometheus/client_golang/prometheus"
```
soltysh, line 7: I'd prefer we re-use k8s component-base like in https://github.com/openshift/library-go/blob/master/pkg/monitor/health/metrics.go since you're get a lot of setup for free from there. Also pls reach out to monitoring team to ensure these metrics are following their requirements.

soltysh (changes requested):

pkg/security/metrics/prometheus/metrics.go
atiratree: nice, I have switched to component- base logic and it looks much better now

dgrisonnet:
I wonder if this could be replaced by the `kube_namespace_annotations` from kube-state-metrics v2.2.0. We are currently only using v2.1.1 in OpenShift, but we will most likely update it to v2.2+ in 4.10 and we could allowlist the scc annotation to expose its value in the `kube_namespace_annotations` metric. @atiratree do you think that metric would match your needs?

atiratree:
> I wonder if this could be replaced by the `kube_namespace_annotations` from kube-state-metrics v2.2.0. We are currently only using v2.1.1 in OpenShift, but we will most likely update it to v2.2+ in 4.10 and we could allowlist the scc annotation to expose its value in the `kube_namespace_annotations` metric. @atiratree do you think that metric would match your needs?

Thanks for letting me know of this interesting capability.

One issue with this approach is that we are not measuring the annotations per se as this is an implementation detail. We want to expose  the state of processing by cpc controller (Pending vs Resolved)).
- the Pending would not be exposed by the kube-state-metrics and would have to be computed (less user friendly)
- we might want to add a Failure state in the future
- the implementation might change as well

Although on the other hand, it would lower the code complexity here if we decided to go with kube-state-metrics.

@soltysh what is your opinion on this?

dgrisonnet:
Can't you infer the scc processing state from the annotation value? I am not too familiar with this feature, but that's what I am gathering from your current implementation. If that's the case, then we can have this information from kube-state-metrics.

Then what you could do is to create recording rules in your operator to match your needs. For example:
```
- record:  openshift:namespace_scc_allocation_resolved:sum
  expr: 'sum(kube_namespace_annotations{openshift.io/sa.scc.uid-range!=""})'
- record:  openshift:namespace_scc_allocation_pending:sum
  expr: 'sum(kube_namespace_annotations{openshift.io/sa.scc.uid-range=""})'
```

atiratree:
+1, that could work in case we don't want to include the errors

pkg/security/metrics/prometheus/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package prometheus contains code for injecting namespace allocation related metrics
```
soltysh, line 1: Nit: just `pkg/security/metrics` for this package name like we do that elsewhere.

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -98,7 +101,18 @@ func NewNamespaceSCCAllocationController(namespaceInformer corev1informers.Names
 		10*time.Minute,
 	)
 
-	return factory.New().ResyncSchedule(resyncPeriod).WithBareInformers(namespaceInformer.Informer()).WithSyncContext(syncContext).WithSync(c.sync).
+	return factory.New().
+		ResyncSchedule(resyncPeriod).
+		WithBareInformers(namespaceInformer.Informer()).
+		WithSyncContext(syncContext).
+		WithSync(c.sync).
+		WithPostStartHooks(func(ctx context.Context, syncContext factory.SyncContext) error {
```
soltysh, line 109: Any particular reason for using post-start hook instead of directly registering metrics in https://github.com/openshift/cluster-policy-controller/blob/master/pkg/cmd/cluster-policy-controller/policy_controller.go#L18 similarly to how we do that for example in https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L358 ?

pkg/security/metrics/prometheus/metrics.go
soltysh: Yes, definitely.

soltysh:
> @soltysh what is your opinion on this?

@atiratree suggestion from @dgrisonnet are on point, can you figure out what's the current state of it and how we can expose that?

pkg/security/metrics/prometheus/doc.go
atiratree: done

pkg/security/controller/namespace_scc_allocation_controller.go
atiratree: I have chosen a different approach now

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -286,11 +295,14 @@ func (c *NamespaceSCCAllocationController) Repair(ctx context.Context, syncCtx f
...
+
+	namespacesProcessed := 0
 	for _, ns := range nsList {
 		value, ok := ns.Annotations[securityv1.UIDRangeAnnotation]
 		if !ok {
 			continue
 		}
+		namespacesProcessed++
 		block, err := uid.ParseBlock(value)
 		if err != nil {
```
atiratree, line 307: do we care about namespaces with invalid blocks? Should we publish these error in form of metrics or other form?

atiratree (commented):
@soltysh I have added metrics for computing percentage of uids allocated. It should be easy to write Alerts for these now.

atiratree:
> Can't you infer the scc processing state from the annotation value? I am not too familiar with this feature, but that's what I am gathering from your current implementation. If that's the case, then we can have this information from kube-state-metrics.

@dgrisonnet 
I have tried these, but could not get these specific annotations in my cluster with prometheus. Nevertheless I am leaning towards more complex approach now, so I will skip these.

dgrisonnet:
These metrics are gated so you will need to allow them manually via kube-state-metrics `--metric-annotations-allowlist` flag. The current downstream configuration resides here: https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/components/kube-state-metrics.libsonnet#L142. We don't have any annotation metrics allowed yet, but you could add the one for namespaces with:
```
--metric-annotations-allowlist=namespaces=[*]
```

atiratree:
Thanks for the info

atiratree:
- there is an ongoing issue with cpc ports not being opened after cluster upgrades
- needs tests for new metrics

/hold until this is  resolved

pkg/security/controller/namespace_scc_allocation_controller.go
soltysh: It would be nice to at least warn about these, either in logs or through some kind of annotation. Can be done in a followup.

pkg/cmd/controller/security.go
```diff
@@ -29,13 +32,20 @@ func RunNamespaceSecurityAllocationController(ctx context.Context, controllerCtx
 		return true, err
 	}
 
+	metrics, err := sccmetrics.NewNamespaceSCCAllocMetrics()
```
soltysh, line 35: If `NamespaceSCCAllocationController` is the only consumer of these metrics, I'm inclined to register them directly inside that controller, and use `MustRegister` which will panic if this doesn't work. That's what we do in k8s controllers, see https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/cronjob_controllerv2.go#L122

pkg/security/controller/namespace_scc_allocation_controller.go
atiratree: +1, I will make a followup PR for it

pkg/cmd/controller/security.go
atiratree: I have moved them inside the controller, but kept the NamespaceSCCAllocationMetrics interface for easier testing

atiratree:
/reopen

this is only blocked on the network traffic propagation

PR #75, opened 2022-03-23
Bug 2086519: Introduce Pod Security Admission Label Synchronization controller

stlaz:
This PR introduced a controller that implements the design from https://github.com/openshift/enhancements/pull/1010

So far this is just a draft as we need to introduce optimizations to only ever consider rolebindings/roles that matter in terms of SCC access to prevent unnecessary number of calculations in case of a big number of Roles/SCCs/both.

/assign @s-urbaniak 
/cc @openshift/openshift-team-auth-maintainers @deads2k


stlaz:
/test all
just for the fun of it

stlaz:
/hold
just in case a random person comes and lgtms, I want at least two pair of eyes on it

stlaz:
/retest
let's see whether go 1.18 builders were finally added

stlaz:
Note to reviewers: the controller won't be able to do anything until https://github.com/openshift/cluster-kube-controller-manager-operator/pull/616 gets merged so you are only reviewing the code.

Once we are fine with the concept, we can merge the code and run CI on it again.

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,440 @@
...
+	rolesSynced        cache.InformerSynced
+	roleBindingsSynced cache.InformerSynced
+
+	usefulRolesLock sync.Mutex
+	usefulRoles     map[string][]string
+}
+
+// role and clusterrolebinding object for generic handling, assumes one and
+// at most one of role/clusterrole is always non-nil
+type roleBindingObj struct {
```
deads2k, line 51: My preference, given usage, this seems more like

```
type bindingInterface interface{
  RoleRef() rbacv1.RoleRef
  Subjects()
  Namespace()
}
```

With this struct as the impl. Otherwise I fear direct access.

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,440 @@
...
+// of strings to create an index for the SA or SA group.
+func BySAIndexKeys(obj interface{}) ([]string, error) {
+	roleBinding, err := newRoleBindingObj(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	serviceAccounts := []string{}
+	for _, subject := range roleBinding.Subjects() {
+		if subject.APIGroup == "" && subject.Kind == "ServiceAccount" {
```
deads2k, line 158: minor: I prefer switch/case and summarizing someting like

```go
IsGroupBinding := subject.APIGroup == rbacv1.GroupName && subject.Kind == "Group"
isAllServiceAccount := subject.Name == serviceaccount.AllServiceAccountsGroup
isNamespacedServiceAccountGroup := strings.HasPrefix(subject.Name, serviceaccount.ServiceAccountGroupPrefix))
```

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -0,0 +1,285 @@
...
+		sccLister:                sccInformer.Lister(),
+
+		nsLabelSelector: controlledNamespacesLabelSelector,
+
+		saToSCCsCache: NewSAToSCCCache(rbacInformers, sccInformer),
+	}
+
+	return factory.New().
+		WithSync(c.sync).
+		WithFilteredEventsInformersQueueKeysFunc(
```
deads2k, line 101: do you guys prefer the way these factor out while using this factory thing or do you prefer "normal" construction?

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,440 @@
...
+	return allowedSCCs, nil
+}
+
+// getRoleFromRoleRef tries to retrieve the role or clusterrole from roleRef.
+func (c *SAToSCCCache) getRoleFromRoleRef(ns string, roleRef rbacv1.RoleRef) (*roleObj, error) {
+	var role interface{}
+	var retryErr error
+	switch kind := roleRef.Kind; kind {
+	case "Role":
+		retryErr = retry.OnError(retry.DefaultBackoff, apierrors.IsNotFound, func() error {
```
deads2k, line 260: This looks questionable.  I would instead rely on being retriggered.  otherwise a set of bad rolebindings can really mess you up.

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,440 @@
...
+// to SCCs and reinitializes it by pulling all cluster-/roles and SCCs and
+// reevaluating them.
+// This should be used rather rarely as there are many computations involved
+// when assessing all the present cluster-/role rules.
+func (c *SAToSCCCache) ReinitializeRoleCache() error {
+	// rewriting the whole cache, lock is needed
+	c.usefulRolesLock.Lock()
+	defer c.usefulRolesLock.Unlock()
+
+	c.usefulRoles = map[string][]string{}
```
deads2k, line 342: I'm not seeing where this is updated over time as roles are updated.

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,440 @@
...
+				subject.Name == user.AllAuthenticated ||
+				strings.HasPrefix(subject.Name, serviceaccount.ServiceAccountGroupPrefix)) {
+			serviceAccounts = append(serviceAccounts, subject.Name)
+		}
+	}
+
+	return serviceAccounts, nil
+}
+
+func NewSAToSCCCache(rbacInformers rbacv1informers.Interface, sccInfomer securityv1informers.SecurityContextConstraintsInformer) *SAToSCCCache {
```
deads2k, line 171: This looks like a potential scale problem to me.  The for loops in here are pretty deeply nested.  If this becomes a bottleneck, how would you make it an order of magnitude faster.  The information is present to fully invert this logic.

pkg/psalabelsyncer/sccrolecache.go
stlaz: Hmm, previously I had it tied to events that come for the controller but it's true that while thinking of it as a separate logical object, it should add its own event handlers to the informers.

pkg/psalabelsyncer/sccrolecache.go
stlaz: :+1:

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: I think the factoring is making it easier to observe what the expectations of the controller might be. Although I haven't written many "normal" kube-controllers, perhaps there would be a way to keep the inputs well readable, too.

pkg/psalabelsyncer/sccrolecache.go
stlaz: I tried, it turns out to be quite a lot of string comparisons that may be unnecessary if I wanted to keep summaries like that. I can keep a switch-case for a) direct SAs; b) groups but the rest I'd keep the same.

pkg/psalabelsyncer/sccrolecache.go
stlaz: I implemented it so that the informer events for roles and SCCs are handled by the cache, and the cache is capable of adding keys to an external queue.

pkg/psalabelsyncer/sccrolecache.go
stlaz: The logic expects minimal changes in terms of SCC creation, which is currently the most expensive operation as the role cache gets reinitialized. That also means that the initialization of the cache on the start is going to be quite expensive - each SCC addition means `(n_SCCs * (n_roles * n_role_rules))` complexity. This could perhaps be improved by the cache dealing with only the SCC that was changed, and therefore not reinitializing the whole cache.

As for the roles and clusterroles - these are handled one-by-one and cached. The computation complexity there is not great - `(n_SCCs * n_role_rules)` - but I think that in this case it's either us or the kube-apiserver doing the computation.

As for the memory complexity, the role cache is just a `string->[]strings` where the string slice are the names of the SCCs. This means that the cache will of course grow with role and SCC numbers, but the number of SCCs should hopefully never be too great and so memory growth here should be rather linear.

stlaz:
The last commit brings optimization of the SCCs as discussed in https://github.com/openshift/cluster-policy-controller/pull/75#discussion_r865741000. It seems that, for some reason, this is blocking removal of the openshift finalizer from namespaces, I'll have to check why that could be.

pkg/psalabelsyncer/sccrolecache.go
stlaz: The cost for adding an SCC has been improved to `(n_roles * n_role_rules)`

stlaz: @s-urbaniak proposed to write a benchmark test as a follow-up

stlaz:
/test all
the RBAC PR merged, let's see how this actually works

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -0,0 +1,343 @@
...
+			serviceAccountInformer.Informer(),
+		).
+		ToController(
+			controllerName,
+			eventRecorder.WithComponentSuffix(controllerName),
+		), nil
+}
+
+func (c *PodSecurityAdmissionLabelSynchronizationController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	const errFmt = "failed to synchronize namespace %q: %v"
```
s-urbaniak, line 141: s/%v/%w

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+func (c *SAToSCCCache) handleSCCAdded(obj interface{}) {
+	scc, ok := obj.(*securityv1.SecurityContextConstraints)
+	if !ok {
+		klog.Errorf("expected object to be of type SecurityContextConstraints but it is %T", obj)
+		return
+	}
+
+	roles, err := c.roleLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to handle %q SCC addition: %v", scc.Name, err)
```
s-urbaniak, line 283: s/%v/%w

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+		return
+	}
+
+	clusterRoles, err := c.clusterRoleLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to handle %q SCC addition: %v", scc.Name, err)
```
s-urbaniak, line 289: s/%v/%w

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+
+		idxRemove := -1
+		for idx, cachedSCC := range sccs {
+			if sccName == cachedSCC {
+				idxRemove = idx
+				break
+			}
+		}
+		// remove the SCC from cache if it got removed
+		if idxRemove != -1 {
```
s-urbaniak, line 337: complete stylistic nit: I believe conditions like this read better if expressed as `idxRemove < 0` but consider this just as a comment :-)

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+// to run pods in its namespace.
+// It expects the serviceAccount name in the system:serviceaccount:<ns>:<name> form.
+func (c *SAToSCCCache) SCCsFor(serviceAccount *corev1.ServiceAccount) (sets.String, error) {
+	saUserInfo := serviceaccount.UserInfo(
+		serviceAccount.Namespace,
+		serviceAccount.Name,
+		string(serviceAccount.UID),
+	)
+
+	// realSAUserInfo adds the "system:authenticated" group to SA UserInfo groups
```
s-urbaniak, line 356: why? (add the why to the comment as well)

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+	var err error
+	var role interface{}
+	switch kind := roleRef.Kind; kind {
+	case "Role":
+		role, err = c.roleLister.Roles(ns).Get(roleRef.Name)
+		if err != nil {
+			return nil, err
+		}
+
+	case "ClusterRole":
```
s-urbaniak, line 427: can we reference constants for this?

pkg/psalabelsyncer/sccrolecache.go
```diff
@@ -0,0 +1,550 @@
...
+	}
+
+	c.usefulRolesLock.Lock()
+	defer c.usefulRolesLock.Unlock()
+
+	return len(c.usefulRoles[usefulRolesKey(role)]) != 0
+}
+
+func (c *SAToSCCCache) syncRoleCacheSingleSCCAdded(role RoleInterface, rules []rbacv1.PolicyRule, scc *securityv1.SecurityContextConstraints) {
+	if c.usefulRolesLock.TryLock() {
```
s-urbaniak, line 463: I am sceptical about this. Are you worried about deadlocks? If so we should rather proof they can occur and prevent them. As per https://pkg.go.dev/sync#Mutex.TryLock:

> Note that while correct uses of TryLock do exist, they are rare, and use of TryLock is often a sign of a deeper problem in a particular use of mutexes.

s-urbaniak:
from my side this is close to lgtm from my side: generally two comments:

- Please fix the error handling and a consistent usage of %w
- I am questioning we need TryLock for the mutex. If it is really needed we should document very well why. If it is to prevent deadlocks we should find the corner cases now and fix them in the first place.

pkg/psalabelsyncer/sccrolecache.go
stlaz: this is a log, it does not get the %w

pkg/psalabelsyncer/sccrolecache.go
stlaz: this is a log, it does not get the %w

pkg/psalabelsyncer/sccrolecache.go
stlaz: the `idxRemove` is getting defined a few lines above with `-1` being the default value and so this condition is only checking whether it changed so I'd say this exact non-equals is a bit more accurate

pkg/psalabelsyncer/sccrolecache.go
stlaz: not that I am aware

pkg/psalabelsyncer/sccrolecache.go
stlaz: This is not about deadlocks. It is a preventive measure for the programmer using the function without a lock.

s-urbaniak: as discussed OOB this is not enough of a justification, xref https://github.com/golang/go/issues/45435#issuecomment-832910477 and https://github.com/golang/go/issues/45435#issuecomment-841028093

s-urbaniak:
@stlaz this looks much better now 👍 I still see quite low unit test coverage for `sccrolecache.go`, it is ~50% as per https://github.com/openshift/cluster-policy-controller/pull/75/commits/b3a5b519e0c6a04e5c446b26c18be2782d37f887.

more specifically:

1. `sccrolecache.go`
- add unit tests for `getRoleFromRoleRef`, `IsRoleBindingRelevant`, `handleRoleModified` (and thus `syncRoleCache`) next to the existing `SCCsFor`
- execute the calls on the above functions from concurrently running go-routines
- ensure that the above test is executed using `go test -race`

2. `podsecurity_label_sync_controller.go`: add unit tests, it's coverage is 0%
3. `scctopsamapping.go`: add unit tests, it's coverage is 0%

stlaz:
/retest
we merged a fix that omits e2e-test namespaces from sync

stlaz:
/retest
CI payloads with the fix should finally be available

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -35,6 +35,78 @@ const (
...
 )
 
+var systemNSSyncExemptions = sets.NewString(
+	// kube-specific system namespaces
+	"default",
+	"kube-node-lease",
+	"kube-public",
+	"kube-system",
+
+	// openshift payloand namespaces
```
s-urbaniak, line 45: s/payloand/payload

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -35,6 +35,78 @@ const (
...
+	"openshift-oauth-apiserver",
+	"openshift-openstack-infra",
+	"openshift-operator-lifecycle-manager",
+	"openshift-operators",
+	"openshift-ovirt-infra",
+	"openshift-sdn",
+	"openshift-service-ca",
+	"openshift-service-ca-operator",
+	"openshift-user-workload-monitoring",
+	"openshift-vsphere-infra",
```
s-urbaniak, line 107: could you describe how you retrieved that list (for further maintanance)

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -301,11 +383,18 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) checkNSControlled(n
 
 func checkNSControlled(ns *corev1.Namespace) bool {
```
s-urbaniak, line 384: i just realized this, to stay go idiomatic (and easier to understand), call this `isNSControlled` and add a godoc (despite not being public this is a very central place of logic).

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -301,11 +383,18 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) checkNSControlled(n
...
 
+	// while "openshift-" namespaces should be considered controller, there are some
+	// edge cases where users can also create them. Consider these a special case
+	// and delegate the decision to sync on the user who should know what they are
+	// doing when creating a NS that appears to be system-controlled.
+	if strings.HasPrefix(nsName, "openshift-") {
+		return ns.Labels[labelSyncControlLabel] == "true"
+	}
+
 	if ns.Labels[labelSyncControlLabel] != "false" {
```
s-urbaniak, line 398: The current logic would consider a namespace `foo` labelled `security.openshift.io/scc.podSecurityLabelSync=meow` to be controlled but a namespace `openshift-foo` not.

stlaz: this is fine

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -301,11 +383,18 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) checkNSControlled(n
 
 func checkNSControlled(ns *corev1.Namespace) bool {
 	nsName := ns.Name
-	isSystemNS := strings.HasPrefix(nsName, "openshift-") || nsName == "openshift" || nsName == "kube-system"
-	if isSystemNS {
+	if systemNSSyncExemptions.Has(nsName) {
 		return false
 	}
 
+	// while "openshift-" namespaces should be considered controller, there are some
```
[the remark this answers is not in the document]
stlaz: it's a typo, it was supposed to be `considered controlled`

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
s-urbaniak: logically, yes, but the interpretation of `true`/`false` values is still inconsistent between the cases. the logic per se is fine, agreed.

s-urbaniak: otoh, these are explicit values we're asking for here, so let's leave it.

s-urbaniak:
(sorry, lgmt'ed the wrong PR

stlaz:
rebased on current master w/ kube 1.24

stlaz:
A CI payload with the required o/origin fixes has not been promoted yet :slightly_frowning_face:

stlaz:
/retest-required
infra

PR #85, opened 2022-08-20
Factor our ns exemptions to make them consumable by other components

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

#### Description
Factors out list of exempt namespaces into its own package, which surfaces a function to check whether a namespace is exempt from sync'ing or not

#### Motivation
The OLM team is creating a process to enable sync'ing on non-payload openshift-* namespaces that contain operators (ClusterServiceVersion resource). It will also ignore exempt namespaces. This change makes it so that the OLM team's process can draw a dependency on the label syncer and user it as a single source of truth for the exemption list.

#### Alternatives
Instead of moving the list to its own package, we could just surface the function from within the `psalabelsyncer` package itself. I felt putting it in its own package was better as it hides the implementation and keeps the list safe from orthogonal changes (insertions, deletions, etc.).

### DIFF
--- a/pkg/psalabelsyncer/nsexemptions/nsexemptions.go
+++ b/pkg/psalabelsyncer/nsexemptions/nsexemptions.go
@@ -1,0 +1,85 @@
+package nsexemptions
+
+import "k8s.io/apimachinery/pkg/util/sets"
+
+// systemNSSyncExemptions is the list of namespaces deployed by an OpenShift install
+// payload, as retrieved by listing the namespaces after a successful installation
+// IMPORTANT: The Namespace openshift-operators must be an exception to this rule
+// since it is used by OCP/OLM users to install their Operator bundle solutions.
+var systemNSSyncExemptions = sets.NewString(
+	// kube-specific system namespaces
+	"default",
+	"kube-node-lease",
+	"kube-public",
+	"kube-system",
+
+	// openshift payload namespaces
+	"openshift",
+	"openshift-apiserver",
+	"openshift-apiserver-operator",
+	"openshift-authentication",
+	"openshift-authentication-operator",
+	"openshift-cloud-controller-manager",
+	"openshift-cloud-controller-manager-operator",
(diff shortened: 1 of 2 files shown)


pkg/psalabelsyncer/nsexemptions/nsexemptions.go
```diff
@@ -0,0 +1,82 @@
...
+	"openshift-operator-lifecycle-manager",
+	"openshift-ovirt-infra",
+	"openshift-sdn",
+	"openshift-service-ca",
+	"openshift-service-ca-operator",
+	"openshift-user-workload-monitoring",
+	"openshift-vsphere-infra",
+)
+
+func IsNamespacePSALabelSyncExempted(namespace string) bool {
```
deads2k, line 80: if you're going to expose this, you need to indicate that is is not accurate unless the cluster version exactly matches the level vendored into the thing using it, which  usually doesn't happen.

perdasilva: Great point. I'll add the target OCP version to the method name itself. Then it's absolutely clear to the consumer.

pkg/psalabelsyncer/nsexemptions/nsexemptions.go
```diff
@@ -0,0 +1,84 @@
...
+	"openshift-ovirt-infra",
+	"openshift-sdn",
+	"openshift-service-ca",
+	"openshift-service-ca-operator",
+	"openshift-user-workload-monitoring",
+	"openshift-vsphere-infra",
+)
+
+// IsNamespacePSALabelSyncExemptedInOCP412 returns true if the given namespace should be exempted from
+// PSA label sync'ing in OCP 4.12
```
deads2k, line 81: This won't age well.  It's `IsNamespacePSALabelSyncExemptedInWhateverOCPLevelYouVendored`

PR #87, opened 2022-08-23
Bump Kubernetes module dependencies to 1.25.

benluddy:


benluddy:
/hold

Will update if 1.25.0 is tagged today.

PR #94, opened 2023-01-11
OCPBUGS-723: clusterquotareconciliation: do not sync quota monitor cache with no monitors registered

ingvagabund:
Empty list of resources is a sign of incorrectness in the cluster. Either the kube-apiserver incorrectly returns an empty list with no error. Or, the client-code responsible for processing response from the discovery is incorrectly interpretting the server error.

If the list of resources is zero, all current monitors are cleared and removed. Followed by invoking c.quotaMonitor.IsSynced which is always false when there's no monitor running. In which case cache.WaitForCacheSync never returns and loops indefinitely. Never resyncing the resources. Thus, none of the resources mentioned on any ClusterResourceQuota's status are recomputed since there's no resource monitor to provide current resource usage.

How to reproduce the issue:
1. create a CRQ:
    ```yaml
    apiVersion: quota.openshift.io/v1
    kind: ClusterResourceQuota
    metadata:
      name: testing-quota
    spec:
      quota:
        hard:
          limits.cpu: 6
          limits.memory: 16G
          requests.cpu: 6
          requests.memory: 16G
          requests.storage: 250G
      selector:
        annotations:
          openshift.io/quota-id: testing-quota
    ```
1. create a testing namespace:
    ```sh
    $ oc create ns testing-namespace
    $ oc annotate ns testing-namespace openshift.io/quota-id=testing-quota
    ```
1. create pods in testing-namespace:
    ```yaml
    apiVersion: apps/v1
    kind: ReplicaSet
    metadata:
      name: frontend
      namespace: testing-namespace
      labels:
        app: guestbook
        tier: frontend
    spec:
      # modify replicas according to your case
      replicas: 120
      selector:
        matchLabels:
          tier: frontend
      template:
        metadata:
          labels:
            tier: frontend
        spec:
          containers:
          - name: php-redis
            image: busybox:1.28
            command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
            resources:
              requests:
                memory: "64Mi"
                cpu: "10m"
              limits:
                memory: "128Mi"
                cpu: "50m"
    ```
1. check the CRQ status:
    ```sh
    $ oc describe ClusterResourceQuota testing-quota
    Name:               testing-quota
    Created:            3 hours ago
    Labels:             <none>
    Annotations:        kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"quota.openshift.io/v1","kind":"ClusterResourceQuota","metadata":{"annotations":{},"name":"testing-quota"},"spec":{"quota":{"hard":{"limits.cpu":6,"limits.memory":"16G","requests.cpu":6,"requests.memory":"16G","requests.storage":"250G"}},"selector":{"annotations":{"openshift.io/quota-id":"testing-quota"}}}}
                
    Namespace Selector: ["testing-namespace"]
    Label Selector:
    AnnotationSelector: map[openshift.io/quota-id:testing-quota]
    Resource            Used            Hard
    --------            ----            ----
    limits.cpu          5950m           6
    limits.memory       15971909632     16G
    requests.cpu        1190m           6
    requests.memory     7985954816      16G
    requests.storage    0               250G

    ```
1. [Hacky] Update the [clusterquotareconciliation controller code](https://github.com/openshift/cluster-policy-controller/blob/f1fe7bf5a097fb85a05d6ed61f4eedea9e71077b/pkg/quota/clusterquotareconciliation/reconciliation_controller.go#L165) to eventually return an empty list of resources with no error. E.g.
    ```go
    newResources, err := resourcequota.GetQuotableResources(discoveryFunc)
    if (condition) {
        newResources = nil
        err = nil
    }
    ```
    Once the condition is met (e.g. 1 minute after the controller startup, at least 2 iteration of the Sync method), setting `newResources` to an empty list will cause the Sync loop to hang indefinitely. After which the created ClusterResourceQuota status no longer gets updated accordingly to the current resource utilization. Everytime a new pod is created (without removing already running pods) and admitted by the kube-apiserver, the resource consumption in the ClusterResourceQuota's status gets increased. Even though the admission plugin code responsible for updating ClusterResourceQuota status increases the consumption, it does not re-computes the overall status. Thus, with the clusterquotareconciliation controller hanging the resource consumption eventually reaches its hard limits. At which point removing already running pods does not decreases the status. After which creation of new pods will be rejected due to exceeded quota which is at its limit.

### DIFF
--- a/pkg/quota/clusterquotareconciliation/reconciliation_controller.go
+++ b/pkg/quota/clusterquotareconciliation/reconciliation_controller.go
@@ -177,6 +177,22 @@ func (c *ClusterQuotaReconcilationController) Sync(discoveryFunc resourcequota.N
 			}
 		}
 
+		// Empty list of resources is a sign of incorrectness in the cluster.
+		// Either the kube-apiserver incorrectly returns an empty list with no error.
+		// Or, the client-code responsible for processing response from the discovery
+		// is incorrectly interpretting the server error.
+		//
+		// If the list of resources is zero, all current monitors are cleared and removed.
+		// Followed by invoking c.quotaMonitor.IsSynced which is always false when there's
+		// no monitor running. In which case cache.WaitForCacheSync never returns
+		// and loops indefinitely. Never resyncing the resources. Thus, none of the resources
+		// mentioned on any ClusterResourceQuota's status are recomputed since there's no
+		// resource monitor to provide current resource usage.
+		if len(newResources) == 0 {
+			klog.V(2).Infof("no resources discovered, skipping resource quota sync")
+			return
+		}
+
 		// Decide whether discovery has reported a change.
 		if reflect.DeepEqual(oldResources, newResources) {
 			klog.V(4).Infof("no resource updates from discovery, skipping resource quota sync")


soltysh (approved):
/lgtm
/approve

Issue #99, opened 2023-02-15
cluster-policy-controller has no template resource informer


lance5890:
* 1. As u have mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1888595 , the template and other resources can not be monitored（the other resources can not be found is normal） ,
* 2.  I have also found this log in my ocp 4.9 envvironment, and i have change the log to --v4; some  log show as follow:

```
I0214 10:49:34.655017       1 resource_quota_monitor.go:181] QuotaMonitor using a shared informer for resource "image.openshift.io/v1, Resource=imagestreams"
I0214 10:49:34.655063       1 resource_quota_monitor.go:248] quota synced monitors; added 1, kept 0, removed 0

I0214 10:49:34.696109       1 resource_quota_monitor.go:181] QuotaMonitor using a shared informer for resource "apps.openshift.io/v1, Resource=deploymentconfigs"
I0214 10:49:34.696162       1 resource_quota_monitor.go:233] QuotaMonitor created object count evaluator for deploymentconfigs.apps.openshift.io

I0214 10:49:34.696607       1 resource_quota_monitor.go:181] QuotaMonitor using a shared informer for resource "build.openshift.io/v1, Resource=buildconfigs"
I0214 10:49:34.696624       1 resource_quota_monitor.go:233] QuotaMonitor created object count evaluator for buildconfigs.build.openshift.io
I0214 10:49:34.696649       1 resource_quota_monitor.go:181] QuotaMonitor using a shared informer for resource "route.openshift.io/v1, Resource=routes"
I0214 10:49:34.696668       1 resource_quota_monitor.go:233] QuotaMonitor created object count evaluator for routes.route.openshift.io

I0214 10:49:34.697000       1 interface.go:71] Couldn't find informer for template.openshift.io/v1, Resource=templates
I0214 10:49:34.697006       1 resource_quota_monitor.go:185] QuotaMonitor unable to use a shared informer for resource "template.openshift.io/v1, Resource=templates": no informer found for template.openshift.io/v1, Resource=templates

I0214 10:49:34.698047       1 resource_quota_monitor.go:181] QuotaMonitor using a shared informer for resource "build.openshift.io/v1, Resource=builds"
I0214 10:49:34.698073       1 resource_quota_monitor.go:233] QuotaMonitor created object count evaluator for builds.build.openshift.io

[...]

PR #100, opened 2023-02-15
OCPBUGS-8271: external template and route Informer

lance5890:
1. external template and route Informer, I think this is caused by commit  https://github.com/openshift/cluster-policy-controller/commit/ccd7b3e29db2c01a73dd42c14096b823ad4d8072 and some other similar commit

2. log and analysis show in this issue:
[# issue 99  cluster-policy-controller log err](https://github.com/openshift/cluster-policy-controller/issues/99)

### DIFF
--- a/pkg/client/genericinformers/interface.go
+++ b/pkg/client/genericinformers/interface.go
@@ -3,7 +3,6 @@ package genericinformers
 import (
 	"k8s.io/klog/v2"
 
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/informers"
 )
@@ -12,19 +11,6 @@ type GenericResourceInformer interface {
 	ForResource(resource schema.GroupVersionResource) (informers.GenericInformer, error)
 	Start(stopCh <-chan struct{})
 }
-
-// GenericInternalResourceInformerFunc will return an internal informer for any resource matching
-// its group resource, instead of the external version. Only valid for use where the type is accessed
-// via generic interfaces, such as the garbage collector with ObjectMeta.
-type GenericInternalResourceInformerFunc func(resource schema.GroupVersionResource) (informers.GenericInformer, error)
-
-func (fn GenericInternalResourceInformerFunc) ForResource(resource schema.GroupVersionResource) (informers.GenericInformer, error) {
-	resource.Version = runtime.APIVersionInternal
-	return fn(resource)
-}
-
-// this is a temporary condition until we rewrite enough of generation to auto-conform to the required interface and no longer need the internal version shim
-func (fn GenericInternalResourceInformerFunc) Start(stopCh <-chan struct{}) {}
 
 // genericResourceInformerFunc will handle a cast to a matching type
 type GenericResourceInformerFunc func(resource schema.GroupVersionResource) (informers.GenericInformer, error)
--- a/pkg/cmd/controller/interfaces.go
+++ b/pkg/cmd/controller/interfaces.go
@@ -141,13 +141,13 @@ func (c *EnhancedControllerContext) ToGenericInformer() genericinformers.Generic
 		genericinformers.GenericResourceInformerFunc(func(resource schema.GroupVersionResource) (informers.GenericInformer, error) {
 			return c.ImageInformers.ForResource(resource)
 		}),
-		genericinformers.GenericInternalResourceInformerFunc(func(resource schema.GroupVersionResource) (informers.GenericInformer, error) {
+		genericinformers.GenericResourceInformerFunc(func(resource schema.GroupVersionResource) (informers.GenericInformer, error) {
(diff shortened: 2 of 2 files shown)


lance5890:
@deads2k @soltysh

lance5890:
@atiratree @mfojtik

lance5890:
Anyone who has any idea? I have modify the cluster-policy-controller code and rebuild the image, and stop the cvo，and change the release image, it does work

lance5890:
@ingvagabund

stlaz:
I am missing context here (even after looking into the issue). What's not working?

lance5890:
> I am missing context here (even after looking into the issue). What's not working?

The cluster-policy-controller log show "unable to use a shared informer for resource "template.openshift.io/v1, Resource=templates": no informer found for template.openshift.io/v1, Resource=templates" ; but when I change the  GenericInternalResourceInformerFunc to GenericResourceInformerFunc, and I replace the new image into the release, the log will never show again

ingvagabund:
This looks like a reasonable fix without the knowledge why the internal version was used in the first place. Nevertheless, the fix needs at least a unit test to validate it. A test invoking NewControllerContext as the top level entry with a complete list of all resources that are watched. Checking the code this might be challenging.

@lance5890 thanks for the effort. I will need to check the code more to properly review your changes.

lance5890:
> This looks like a reasonable fix without the knowledge why the internal version was used in the first place. Nevertheless, the fix needs at least a unit test to validate it. A test invoking NewControllerContext as the top level entry with a complete list of all resources that are watched. Checking the code this might be challenging.
> 
> @lance5890 thanks for the effort. I will need to check the code more to properly review your changes.

@ingvagabund thanks for your review, that does make sense !

1. I will take free time to study the writing and testing of ut, it will take some time 
2. I have rebuilt the cluster-policy-controller image within my fix, and then stop the cvo, replace the new image into my ocp cluster, is does work

knelasevero:
@lance5890 are you planning to follow up in this issue?

lance5890:
> @lance5890 are you planning to follow up in this issue?

yeah, could u review my PR

knelasevero:
I mean taking care of the test that Jan mentioned before. The PR looks fine with the same reservations as Jan pointed out

lance5890:
> I mean taking care of the test that Jan mentioned before. The PR looks fine with the same reservations as Jan pointed out

i'd like to complete this issue, would u help me to finish this together?

ingvagabund:
The code has been sitting here for a long time. For that it is not easily seen impact of renaming GenericInternalResourceInformerFunc -> GenericResourceInformerFunc. Also, CPC code could use some maintenance phase where we improve the test coverage.

ingvagabund:
```
// GenericInternalResourceInformerFunc will return an internal informer for any resource matching
// its group resource, instead of the external version. Only valid for use where the type is accessed
// via generic interfaces, such as the garbage collector with ObjectMeta.
type GenericInternalResourceInformerFunc func(resource schema.GroupVersionResource) (informers.GenericInformer, error)
```
vs.
```
// genericResourceInformerFunc will handle a cast to a matching type
type GenericResourceInformerFunc func(resource schema.GroupVersionResource) (informers.GenericInformer, error)
```

The notable difference: `GenericInternalResourceInformerFunc only valid for use where the type is accessed via generic interfaces`. Only applies for Quotas and Templates. Also
```
// this is a temporary condition until we rewrite enough of generation to auto-conform to the required interface and no longer need the internal version shim
func (fn GenericInternalResourceInformerFunc) Start(stopCh <-chan struct{}) {}
```

ingvagabund:
Comparing https://github.com/openshift/client-go/blob/master/quota/informers/externalversions/generic.go and https://github.com/openshift/client-go/blob/master/image/informers/externalversions/generic.go and other files on common both `externalversions` are generated using the same generators.

ingvagabund:
Checking https://github.com/openshift/origin/commit/555f87c5530c8397fa86b8e45cfaafdb1d26c515#diff-9a553c47b0996d1d4fb15a1ea194e97c15abbe737b9fda6f05eee890ae7a4045L154 it seems `GenericInternalResourceInformerFunc` -> `GenericResourceInformerFunc` change was just forgotten :)

ingvagabund:
@lance5890 the internal version was used long time ago before there were versioned clientset. Can you also remove definition of `GenericInternalResourceInformerFunc` as well? Both quota and template resources import externalversion informers already. The changes looks good.

lance5890:
> @lance5890 the internal version was used long time ago before there were versioned clientset. Can you also remove definition of `GenericInternalResourceInformerFunc` as well? Both quota and template resources import externalversion informers already. The changes looks good.

ok I will get a closer look at this, and repaire what u said

lance5890:
> /approve

could we cherry-pick this to 4.12?

atiratree:
+1, I think it makes to backport this
/cherry-pick release-4.13

PR #102, opened 2023-02-17
OCPBUGS-7705: [release-4.12] update dependencies to point to v0.25.0

perdasilva:
This is a backport of https://github.com/openshift/cluster-policy-controller/pull/101 and https://github.com/openshift/cluster-policy-controller/pull/103 - it seems the operator-manager dependency is breaking ART


bparees:
does not actually change any vendored code, so nothing for QE to validate, marking cherry-pick-approved

PR #123, opened 2023-07-19
psalabelsyncer: synchronize also audit and warn labels when enforcing

tiraboschi:
Currently the EnforcingPodSecurityAdmissionLabelSynchronizationController is setting only the `pod-security.kubernetes.io/enforce` label ignoring `pod-security.kubernetes.io/warn` and `pod-security.kubernetes.io/audit` ones (it's eventually simply removing leftovers from 4.11). This means that we can end in situations where the POD is actually admitted at `privileged` level altough it's still audited and it will still cause a user user-facing warning.

Let's sync also `pod-security.kubernetes.io/audit` and `pod-security.kubernetes.io/warn` at the same level when setting `pod-security.kubernetes.io/enforce` to get a consistent behaviour.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2223948

### DIFF
--- a/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
+++ b/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
@@ -255,29 +255,19 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 	var changed bool
 
 	if c.shouldEnforce {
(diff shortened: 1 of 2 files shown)


tiraboschi:
@stlaz @deads2k FYI

tiraboschi:
@stlaz @deads2k ping

stlaz:
Hello,
Thank you for your contribution. We reworked how the controller operates in #127, and it now syncs all the labels as this PR proposes.

/close

PR #115, opened 2023-05-22
OCPBUGS-13649: fix ClusterResourceQuotas to work for all api resources including custom resources

atiratree:

PR #124, opened 2023-07-20
OCPBUGS-15568: Add timeout into cache sync wait to prevent hanging forever

ardaguclu:
Previously, cluster quota reconciliation used small subset of informers and that worked without any 
problem with the one important downside that only these small subset of resources were supported
by `clusterresourcequota` unlike to the builtin `resourcequota`.

After https://github.com/openshift/cluster-policy-controller/pull/115, cluster quota started using 
genericinformers(more specifically metadatainformers) rather than pre-defined static resource based informers to  support all other resources including CRDs.

But that brings about hanging problem as defined;

* cluster quota monitor acquires lock to start informers for all the newly created resources
* workers in https://github.com/openshift/cluster-policy-controller/blob/2a9b7b4daa7f4b7a5a0a3536426371a78390ed16/pkg/quota/clusterquotareconciliation/reconciliation_controller.go#L298 wait the workerlock being released by quota monitor
* however, due to this https://github.com/kubernetes/kubernetes/issues/71450 issue, quota monitor hangs "forever"
* which leads that controllers in quota reconciliation are locked "forever".

This was fixed in upstream with this commit https://github.com/kubernetes/kubernetes/pull/74747/commits/739df5452a407636a927ac75395784d5bcca9460(and this https://github.com/kubernetes/kubernetes/pull/74747/commits/e5f7af70583b758978c819cd5c0c6a3ddb6fee28) and this PR takes this commit's relevant bits in cluster-policy-controller.

**How we noticed that behavior?**

Our CI test `Cluster resource quota should control resource limits across namespaces` started failing sporadically because quota reconciliation started listening all the resources including the test CRDs created by other tests which resulted in these test CRDs causing creation of informers in quota controller.

### DIFF
--- a/pkg/quota/clusterquotareconciliation/reconciliation_controller.go
+++ b/pkg/quota/clusterquotareconciliation/reconciliation_controller.go
@@ -201,9 +201,7 @@ func (c *ClusterQuotaReconcilationController) Sync(discoveryFunc resourcequota.N
 			return
 		}
 
-		// Something has changed, so track the new state and perform a sync.
-		klog.V(2).Infof("syncing resource quota controller with updated resources from discovery: %v", newResources)
-		oldResources = newResources
+		klog.V(2).Infof("syncing resource quota controller with updated resources from discovery: %s", printDiff(oldResources, newResources))
 
 		// Ensure workers are paused to avoid processing events before informers
 		// have resynced.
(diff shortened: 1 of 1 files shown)


atiratree:
thanks for hunting down the flake and bringing this in line with the upstream fix
/lgtm

ardaguclu:
unrelated failure;
/test e2e-aws

atiratree:
/lgtm cancel

@ardaguclu  we should do this also for the normal resource quota since it should manifest the same problem

https://github.com/openshift/origin/blob/dd46b809c35b4d811aa8bf6fd6e8cf878b19d799/test/extended/quota/resourcequota.go#L93

upstream tests should be covered: https://github.com/kubernetes/kubernetes/blob/fa88c0b7796170eeff5686ae1d7d0f2f3f0df5de/test/e2e/apimachinery/resource_quota.go#L60

atiratree:
^ cancel on a wrong PR (was meant for https://github.com/openshift/origin/pull/28071)
/lgtm

PR #125, opened 2023-07-20
OCPBUGS-15568: Remove debugs logs in workqueuebucket

ardaguclu:
As promised in https://github.com/openshift/cluster-policy-controller/pull/122, this PR removes debug logs
in workerqueuebucket. Only this https://github.com/openshift/cluster-policy-controller/pull/122/files#diff-2c6dda2c88f9fb3af36559ec94c41eebb965f32f37801a57b5f7250ea744982eR221 log stays because upstream 
also has it and it is important while troubleshooting.


atiratree:
we should not be needing  these after https://github.com/openshift/cluster-policy-controller/pull/124 fix
/lgtm

PR #127, opened 2023-08-01
AUTH-413: ps syncer: only sync labels if noone else is managing them

stlaz:
/cc @deads2k 
/cc @ibihim

### DIFF
--- a/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
+++ b/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
@@ -40,8 +40,10 @@ const (
 )
 
 var (
-	enforcementLabels = map[string]string{
+	allPSaLabels = map[string]string{
 		psapi.EnforceLevelLabel: psapi.EnforceVersionLabel,
+		psapi.WarnLevelLabel:    psapi.WarnVersionLabel,
+		psapi.AuditLevelLabel:   psapi.AuditVersionLabel,
 	}
 	loggingLabels = map[string]string{
 		psapi.WarnLevelLabel:  psapi.WarnVersionLabel,
@@ -54,7 +56,7 @@ var (
 // admission namespace label to match the user account privileges in terms of being able
 // to use SCCs
 type PodSecurityAdmissionLabelSynchronizationController struct {
-	shouldEnforce bool
+	syncedLabels map[string]string
 
 	namespaceClient corev1client.NamespaceInterface
 
@@ -77,7 +79,7 @@ func NewEnforcingPodSecurityAdmissionLabelSynchronizationController(
 	eventRecorder events.Recorder,
 ) (factory.Controller, error) {
 	return newPodSecurityAdmissionLabelSynchronizationController(
-		true,
+		allPSaLabels,
 		namespaceClient,
 		namespaceInformer,
 		rbacInformers,
@@ -96,7 +98,7 @@ func NewAdvisingPodSecurityAdmissionLabelSynchronizationController(
 	eventRecorder events.Recorder,
 ) (factory.Controller, error) {
 	return newPodSecurityAdmissionLabelSynchronizationController(
-		false,
+		loggingLabels,
 		namespaceClient,
 		namespaceInformer,
 		rbacInformers,
@@ -107,7 +109,7 @@ func NewAdvisingPodSecurityAdmissionLabelSynchronizationController(
 }
 
 func newPodSecurityAdmissionLabelSynchronizationController(
-	shouldEnforce bool,
+	syncedLabels map[string]string,
 	namespaceClient corev1client.NamespaceInterface,
 	namespaceInformer corev1informers.NamespaceInformer,
 	rbacInformers rbacv1informers.Interface,
@@ -136,7 +138,7 @@ func newPodSecurityAdmissionLabelSynchronizationController(
 
 	syncCtx := factory.NewSyncContext(controllerName, eventRecorder.WithComponentSuffix(controllerName))
 	c := &PodSecurityAdmissionLabelSynchronizationController{
-		shouldEnforce: shouldEnforce,
+		syncedLabels: syncedLabels,
 
 		namespaceClient: namespaceClient,
 
@@ -237,14 +239,12 @@ func forceHistoricalLabelsOwnership(ctx context.Context, nsClient corev1client.N
 
 	cpcOwnedPSaLabels := map[string]string{}
 	// filter out all the labels not owned by this controller
-	for _, labelMap := range []map[string]string{enforcementLabels, loggingLabels} {
-		for labelType, labelVersion := range labelMap {
-			if cpcOwnedLabelKeys.Has(labelType) {
-				cpcOwnedPSaLabels[labelType] = ns.Labels[labelType]
-			}
-			if cpcOwnedLabelKeys.Has(labelVersion) {
-				cpcOwnedPSaLabels[labelVersion] = ns.Labels[labelVersion]
-			}
+	for labelType, labelVersion := range allPSaLabels {
+		if cpcOwnedLabelKeys.Has(labelType) {
+			cpcOwnedPSaLabels[labelType] = ns.Labels[labelType]
+		}
+		if cpcOwnedLabelKeys.Has(labelVersion) {
+			cpcOwnedPSaLabels[labelVersion] = ns.Labels[labelVersion]
 		}
 	}
 
@@ -339,36 +339,45 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 	if err != nil {
 		return fmt.Errorf("ns extraction failed: %w", err)
 	}
-	syncedLabels := enforcementLabels
-	if !c.shouldEnforce {
-		syncedLabels = loggingLabels
+
+	// we must extract the NS in case only some of the labels we're setting need
+	// updating to avoid hotlo
+	nsApplyConfig := corev1apply.Namespace(ns.Name)
+	if err != nil {
+		return fmt.Errorf("failed to extract field ownership for NS %q: %w", ns.Name, err)
 	}
 
-	nsApplyConfig := corev1apply.Namespace(ns.Name)
-	for typeLabel, versionLabel := range syncedLabels {
-		if ns.Labels[typeLabel] != string(psaLevel) || ns.Labels[versionLabel] != currentPSaVersion {
+	var shouldUpdate bool
+	for typeLabel, versionLabel := range c.syncedLabels {
+		if manager := managedNamespaces.getManagerForLabel(typeLabel); len(manager) == 0 || manager == controllerName {
 			nsApplyConfig.WithLabels(map[string]string{
-				typeLabel:    string(psaLevel),
+				typeLabel: string(psaLevel),
+			})
+			if ns.Labels[typeLabel] != string(psaLevel) {
+				shouldUpdate = true
+			}
+		}
+		if manager := managedNamespaces.getManagerForLabel(versionLabel); len(manager) == 0 || manager == controllerName {
+			nsApplyConfig.WithLabels(map[string]string{
 				versionLabel: currentPSaVersion,
 			})
-		}
-	}
-
-	for labelKey := range nsApplyConfig.Labels {
-		if manager := managedNamespaces.getManagerForLabel(labelKey); len(manager) > 0 && manager != controllerName {
-			delete(nsApplyConfig.Labels, labelKey)
-		}
-	}
-
-	if len(nsApplyConfig.Labels) > 0 {
-		_, err := c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
-		if err != nil {
-			if apierrors.IsConflict(err) {
-				klog.V(4).Info("someone else is already managing the PSa labels: %v", err)
-				return nil
+			if ns.Labels[versionLabel] != currentPSaVersion {
+				shouldUpdate = true
 			}
-			return fmt.Errorf("failed to update the namespace: %w", err)
 		}
+	}
+
+	if !shouldUpdate {
+		return nil
+	}
+
+	_, err = c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
+	if err != nil {
+		if apierrors.IsConflict(err) {
+			klog.Warning("someone else is already managing the PSa labels: %v", err)
+			return nil
+		}
+		return fmt.Errorf("failed to update the namespace: %w", err)
 	}
 
 	return nil
--- a/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
+++ b/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/sets"
+	corev1apply "k8s.io/client-go/applyconfigurations/core/v1"
 	corev1informers "k8s.io/client-go/informers/core/v1"
 	rbacv1informers "k8s.io/client-go/informers/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -287,19 +291,19 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
+				nsApplyConfig.Labels[typeLabel] = string(psaLevel)
+				nsApplyConfig.Labels[versionLabel] = currentPSaVersion
 
 			}
 		}
 	}
 
 	if changed {
-		_, err := c.namespaceClient.Update(ctx, nsCopy, metav1.UpdateOptions{})
+		_, err := c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
```
deads2k, line 306: This doesn't do what you think it does.   You should create a new namespace

```
	corev1apply.Namespace("name").WithLabels(map[string]string{
		"one":"val",
		"two":"",
	})
```

That way your field manager only attempts to set what you want.

deads2k, line 306: Now, separately, you've done right by not setting "force".  Now you need to allow for someone else to own these fields.  that means conflict errors are not retried.

So that apierrors.IsConflict() is handled as not an error.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,19 +251,22 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
 
-	nsCopy := ns.DeepCopy()
+	nsApplyConfig, err := corev1apply.ExtractNamespace(ns, controllerName)
+	if err != nil {
+		return fmt.Errorf("failed to convert namespace to its apply configuration form: %w", err)
+	}
 
 	var changed bool
 
 	if c.shouldEnforce {
```
deads2k, line 261: I think it would be easier to debug by checking in metadata.managedFields

```yaml
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          f:hypershift.openshift.io/monitoring: {}
          f:openshift.io/cluster-monitoring: {}
      f:spec: {}
    manager: hypershift
    operation: Apply
    time: "2023-07-26T18:09:23Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:openshift.io/sa.scc.mcs: {}
          f:openshift.io/sa.scc.supplemental-groups: {}
          f:openshift.io/sa.scc.uid-range: {}
        f:labels:
          f:pod-security.kubernetes.io/audit: {}
          f:pod-security.kubernetes.io/audit-version: {}
          f:pod-security.kubernetes.io/warn: {}
          f:pod-security.kubernetes.io/warn-version: {}
    manager: cluster-policy-controller
    operation: Update
    time: "2023-07-26T18:09:23Z"
```

Notice how labels are subdivided by individual name (two owners in that snippet), so you can see if *your* controller set the value.

stlaz: Isn't it better to work with the library rather than our own unstructured handling?

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: The above sets "two" to an empty string. We don't want that.

From what I can see, though, if we use
```
	corev1apply.Namespace("name").WithLabels(map[string]string{
		"one":"val",
	})
```
all our managed labels but "one" will get removed.

We can use that instead, then, that should actually make the code a bit cleaner, hopefully.

stlaz: :exclamation:  NOTE to self:
We cannot use `controllerName` as the manager because we would not be able to update fields from previous release. We must use `cluster-policy-controller` instead.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,57 +262,40 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
-			changed = true
-			if nsCopy.Labels == nil {
-				nsCopy.Labels = map[string]string{}
-			}
-
-			nsCopy.Labels[psapi.EnforceLevelLabel] = string(psaLevel)
-			nsCopy.Labels[psapi.EnforceVersionLabel] = currentPSaVersion
+	managedNamespaces, err := extractNSFieldsPerManager(ns)
+	if err != nil {
+		if err != nil {
```
liouk, line 340: Duplicate `if` statement.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -415,3 +516,56 @@ func controlledNamespacesLabelSelector() (labels.Selector, error) {
...
+}
+
+func managedLabels(fieldsEntry metav1.ManagedFieldsEntry) (sets.Set[string], error) {
+	managedUnstructured := map[string]interface{}{}
+	err := json.Unmarshal(fieldsEntry.FieldsV1.Raw, &managedUnstructured)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal managed fields: %w", err)
+	}
+
+	labels, found, err := unstructured.NestedMap(managedUnstructured, "f:metadata", "f:labels")
```
deads2k, line 556: does this error if there are no labels owned by this particular user?

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -394,12 +389,45 @@ func isNSControlled(ns *corev1.Namespace) bool {
...
+	}
+
+	var labelsOwned, owningAtLeastOneLabel bool
+	for _, labelName := range []string{
+		psapi.EnforceLevelLabel, psapi.EnforceVersionLabel,
+		psapi.WarnLevelLabel, psapi.WarnVersionLabel,
+		psapi.AuditLevelLabel, psapi.AuditVersionLabel,
+	} {
+		if _, ok := ns.Labels[labelName]; ok {
+			if manager := extractedPerManager.getManagerForLabel(labelName); len(manager) > 0 && !(manager == "cluster-policy-controller" || manager == controllerName) {
```
liouk, line 490: This could be more readable as
```go
len(manager) > 0 && manager != "cluster-policy-controller" && manager != controllerName
```

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -415,3 +516,56 @@ func controlledNamespacesLabelSelector() (labels.Selector, error) {
 
 	return labels.NewSelector().Add(*labelRequirement), nil
 }
+
+// extractedNamespaces serves as a cache so that we don't have to re-extract the namespaces
+// for each label. It helps us prevent performance overhead from multiple deserializations.
```
deads2k, line 521: what is it keyed by and what's in the set?

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: This only errors if it encounters an invalid structure in the path (not `map[string]interface{}`). It should never happen.

If no labels are owned, we should get `found = false`

liouk:
/lgtm
/hold

holding for @deads2k to review

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -191,13 +204,85 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) sync(ctx context.Co
...
+	}
+
+	// previously, we were using Update to set the labels, Kube does not consider that as actually owning the fields, even
+	// though it shows up in managedFields and would cause conflicts on value change. Eh. Ugly.
+	//
+	// Writing custom logic that checks which fields are _really_ managed by a manager, caches the unstructured object and then
+	// conditionally removes the label from all those unstructured fields and shoves them back in the proper place
+	// in the object managedFields is tedious, ugly and super error-prone.
+	//
+	// Just remove the fields as the previous owner and quickly readd them as the new one.
```
deads2k, line 277: Rather than do this, could we simply do an Apply with the `force` option set to true?

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,57 +335,40 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
-			nsCopy.Labels[psapi.EnforceVersionLabel] = currentPSaVersion
+	managedNamespaces, err := extractNSFieldsPerManager(ns)
+	if err != nil {
+		if err != nil {
+			return fmt.Errorf("ns extraction failed: %w", err)
 		}
+	}
+	syncedLabels := enforcementLabels
+	if !c.shouldEnforce {
+		syncedLabels = loggingLabels
```
deads2k, line 346: if we are enforcing, shouldn't we check the ownership of all labels?

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,57 +335,40 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
+	}
 
-		// cleanup audit and warn labels from version 4.11
-		// TODO: This can be removed in 4.13 and allow users set these as they wish
```
deads2k, line 269: Are we sure.  Aren't these used for our alerting?  Given the ability to yield to a different client, it seems like we'd want to continue setting these forever.

if we do not set these labels, then our default of "restricted" is used and this results in noise for namespaces that are not restricted because they constantly get warned and (I think), we get alerts that are not accurately indicating when a namespace will violate. 

In fact, is this leading to false positives today?

deads2k: The concerning flow is

1. our default is restricted for audit and warn
2. the namespace is set to privileged
3. we helpfully remove the audit and warn labels
4. a privileged pod is created and fails audit and warn
5. this causes our alert about how many future violates we would have to fail.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,57 +335,40 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
 		}
 	}
 
-	if changed {
-		_, err := c.namespaceClient.Update(ctx, nsCopy, metav1.UpdateOptions{})
+	if len(nsApplyConfig.Labels) > 0 {
+		_, err := c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
 		if err != nil {
+			if apierrors.IsConflict(err) {
+				klog.V(4).Info("someone else is already managing the PSa labels: %v", err)
```
deads2k, line 369: make this a warning I think.  Given the effort, this is pretty unexpected.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,57 +335,40 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
+		if manager := managedNamespaces.getManagerForLabel(labelKey); len(manager) > 0 && manager != controllerName {
+			delete(nsApplyConfig.Labels, labelKey)
 		}
 	}
 
-	if changed {
-		_, err := c.namespaceClient.Update(ctx, nsCopy, metav1.UpdateOptions{})
+	if len(nsApplyConfig.Labels) > 0 {
+		_, err := c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
 		if err != nil {
```
deads2k, line 367: reduce nesting with

```go
if is conflict
  log
  return nil
if err != nil
  return err
```

deads2k:
worthy of an e2e test once merged.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: That's not possible. The field is owned by "cluster-policy-controller + Update". You cannot set the label to an empty value (fails validation). That means that the only action you can do with it is to remove it. But I don't think there is a way to set up a `NamespaceApplyConfiguration` that would express that intention.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: I think you're right and we should set all.

My original thinking back then was we do not set these fields because we're only ever going to be setting enforcement globally, and we will no longer alert at that point. I still think that that's what we should do, however, enforcement can now be enabled via feature flags while the global config stays + since the labels can now be adjusted manually separately, setting all just seems more logical.

I'll fix that.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
stlaz: I'd like to keep the block that contains error cases like this, unless we decide to handle more errors and introduce a switch here

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
liouk: This probably trickled through some commit somewhere as it's still here :)

liouk (changes requested):

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
deads2k: > I'd like to keep the block that contains error cases like this, unless we decide to handle more errors and introduce a switch here

Non-standard with both golang and kube.  Sort of like using `== ""` versus `len(foo) == 0`.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
deads2k: huh

deads2k: > That's not possible. The field is owned by "cluster-policy-controller + Update". You cannot set the label to an empty value (fails validation). That means that the only action you can do with it is to remove it. But I don't think there is a way to set up a `NamespaceApplyConfiguration` that would express that intention

Can't you set it to the current value with a force?

stlaz: If you set it to the current value, the "force" does not seem to matter at all. You will always get double ownership, which will still result in a conflict when you're trying to change it. Some fun corner cases I tested:

1. `cluster-policy-controller` + apply to the same value
    a. the labels is now managed by `cluster-policy-controller + update` and `cluster-policy-controller + apply`
    b. you get the applyconfig and remove the ownership in an apply as the cluster-policy-controller
    c. the field is now again owned only by `cluster-policy-controller + update`
2. `this-controller` + apply  the same value
   a. the label is now managed by `cluster-policy-controller + update` and `this-controller + apply`
   b. try to change the label value as this-controller - you get a conflict

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -250,59 +335,49 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
...
 
-	nsCopy := ns.DeepCopy()
+	managedNamespaces, err := extractNSFieldsPerManager(ns)
+	if err != nil {
+		return fmt.Errorf("ns extraction failed: %w", err)
+	}
 
-	var changed bool
+	// we must extract the NS in case only some of the labels we're setting need
+	// updating to avoid hotlo
```
liouk, line 344: @stlaz just realized that this comment got somehow truncated, just FYI

PR #129, opened 2023-08-22
OCPBUGS-17989: pkg/psalabelsyncer: enforce syncing in case label is set

s-urbaniak:
As discussed with @stlaz this fixes syncing PSa labels if `security.openshift.io/scc.podSecurityLabelSync` is set to `true` by the user.

Currently this case is ommitted if managed fields are set to cluster policy controller.

/cc @deads2k @ibihim @liouk @stlaz

### DIFF
--- a/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
+++ b/pkg/psalabelsyncer/podsecurity_label_sync_controller.go
@@ -348,8 +348,9 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 	}
 
 	var shouldUpdate bool
+	mustSync := ns.Labels[labelSyncControlLabel] == "true"
 	for typeLabel, versionLabel := range c.syncedLabels {
-		if manager := managedNamespaces.getManagerForLabel(typeLabel); len(manager) == 0 || manager == controllerName {
+		if manager := managedNamespaces.getManagerForLabel(typeLabel); len(manager) == 0 || manager == controllerName || mustSync {
 			nsApplyConfig.WithLabels(map[string]string{
 				typeLabel: string(psaLevel),
 			})
@@ -357,7 +358,7 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 				shouldUpdate = true
 			}
 		}
-		if manager := managedNamespaces.getManagerForLabel(versionLabel); len(manager) == 0 || manager == controllerName {
+		if manager := managedNamespaces.getManagerForLabel(versionLabel); len(manager) == 0 || manager == controllerName || mustSync {
 			nsApplyConfig.WithLabels(map[string]string{
 				versionLabel: currentPSaVersion,
 			})
@@ -371,7 +372,7 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 		return nil
 	}
 
-	_, err = c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
+	_, err = c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName, Force: mustSync})
 	if err != nil {
 		if apierrors.IsConflict(err) {
 			klog.Warning("someone else is already managing the PSa labels: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -348,16 +348,17 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 	}
 
 	var shouldUpdate bool
+	mustSync := ns.Labels[labelSyncControlLabel] == "true"
```
ibihim, line 351: We can stop immediately, if we shall not sync:

```suggestion
	mustSync := ns.Labels[labelSyncControlLabel] == "true"

	if mustNotSync := ns.Labels[labelSyncControlLabel] == "false"; mustNotSync {
	    return nil
	}
```

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
```diff
@@ -362,18 +405,21 @@ func TestEnforcingPodSecurityAdmissionLabelSynchronizationController_sync(t *tes
 			"controlled-namespace-previous-enforce-labels/testspecificsa3":                  sets.NewString("scc_restricted"),
 			"controlled-namespace-previous-warn-labels/testspecificsa3":                     sets.NewString("scc_restricted"),
 			"controlled-namespace-previous-enforce-version-different-owner/testspecificsa3": sets.NewString("scc_restricted"),
+			"controlled-namespace-different-owner-and-sync-label-true/testspecificsa3":      sets.NewString("scc_restricted"),
+			"controlled-namespace-different-owner-and-sync-label-unset/testspecificsa3":     sets.NewString("scc_restricted"),
```
ibihim, line 409: I have not enough insights into the test cases, but don't we need to list: `non-controlled-namespace-different-owner-and-sync-label-false`?

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
```diff
@@ -40,6 +40,49 @@ func testNamespaces() []*corev1.Namespace {
...
+				psapi.EnforceVersionLabel:                        "bogus enforce version value",
+				"security.openshift.io/scc.podSecurityLabelSync": "true",
+			},
+			ManagedFields: managedLabelsFields("someone-else",
+				psapi.EnforceLevelLabel,
+				psapi.EnforceVersionLabel,
+			),
+		}},
+		{ObjectMeta: metav1.ObjectMeta{
+			Name:        "non-controlled-namespace-different-owner-and-sync-label-false",
```
ibihim, line 57: What is the meaning of `non-controlled` and `controlled`? The only difference between this and the namespace above is the label sync value, which is mentioned explicitly at the end of the namespace's name.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
s-urbaniak: I would advice against adding a second flow next to `shouldUpdate`. This is adding unnecessary complexity.

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
s-urbaniak: as far as i see this denotes what namespaces are "controlled" or "not controlled" by the policy controller.

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
s-urbaniak: No we don't need to. As this namespace is non-controlled, we don't need a service account fixture.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
s-urbaniak: additionally your suggestion does not pass unit tests.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -371,7 +372,7 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 		return nil
 	}
 
-	_, err = c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName})
+	_, err = c.namespaceClient.Apply(ctx, nsApplyConfig, metav1.ApplyOptions{FieldManager: controllerName, Force: true})
```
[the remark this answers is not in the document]
s-urbaniak: `Force: mustSync` sgtm 👍

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
ibihim: This is strange, but I guess it is due to the state of the unit tests?

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
ibihim: Thx for the explanation.

PR #137, opened 2023-10-18
OCPBUGS-21638: bump(k8s,openshift) to address CVE-2023-44487 

atiratree:


go.mod
```diff
@@ -124,31 +124,29 @@ require (
 )
 
 replace (
```
ingvagabund, line 126: @You can drop the `replace` part

PR #141, opened 2023-10-23
DO NOT MERGE: Proof CVE 4.10

dgrisonnet:
Based on:
- https://github.com/openshift/kubernetes-apiserver/pull/46
- https://github.com/openshift/kubernetes-apimachinery/pull/19


dgrisonnet:
Updated the PR with the code that add the flags. Here is the output:
```sh
./cluster-policy-controller | grep "UnauthenticatedHTTP2DOSMitigation"
                                         UnauthenticatedHTTP2DOSMitigation=true|false (BETA - default=true)
```

dgrisonnet:
The failure seems unrelated, I am restarting.

/retest-required

PR #148, opened 2024-04-26
MON-3866: chore: csr: Add approver for monitoring CSRs issued for metrics-server

machine424:
Also updates the library-go dep to get https://github.com/openshift/library-go/pull/1905.

Currently, cluster-policy-controller  only approves CSR with `prometheus-k8s` as the certificate subject.
With https://github.com/openshift/cluster-monitoring-operator/pull/2536, we need to issue a separate certificate for `metrics-server` SA. (we want to remove the dependency on prometheus' + ease/improve client identification)


pkg/cmd/controller/csr.go
```diff
@@ -24,23 +24,35 @@ func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedContro
...
 		csr.NewLabelFilter(selector),
-		csr.NewServiceAccountApprover(monitoringServiceAccountNamespace, monitoringServiceAccountName, monitoringCertificateSubject),
-		controllerCtx.EventRecorder)
+		csr.NewServiceAccountApprover(monitoringServiceAccountNamespace, monitoringServiceAccountName, monitoringPrometheusAsCertificateSubject),
+		controllerCtx.EventRecorder).Run(ctx, 1)
 
-	go controller.Run(ctx, 1)
+	// For CSRs issued for the metrics-server Pods.
+	go csr.NewCSRApproverController(
+		// TODO: Different name??
```
machine424, line 47: WDYT?

pkg/cmd/controller/csr.go
```diff
@@ -24,23 +24,35 @@ func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedContro
...
+		controllerName,
+		nil,
+		kubeClient.CertificatesV1().CertificateSigningRequests(),
+		controllerCtx.KubernetesInformers.Certificates().V1().CertificateSigningRequests(),
+		csr.NewLabelFilter(selector),
+		// TODO: maybe make it possible in library-go to pass multiple SA??
```
machine424, line 53: WDYT?

pkg/cmd/controller/csr.go
```diff
@@ -24,23 +22,28 @@ func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedContro
...
+		labelsRequirement, err := labels.NewRequirement("metrics.openshift.io/csr.subject", selection.Equals, []string{subject})
+		if err != nil {
+			return true, err
+		}
+		selector.Add(*labelsRequirement)
+	}
+
+	for subject, selector := range subjectSelectors {
+		go csr.NewCSRApproverController(
+			controllerName,
```
machine424, line 39: have different controller names?

pkg/cmd/controller/csr.go
```diff
@@ -24,23 +22,28 @@ func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedContro
 		return true, err
 	}
 
-	selector := labels.NewSelector()
-	labelsRequirement, err := labels.NewRequirement(monitoringLabelKey, selection.Equals, []string{monitoringLabelValue})
-	if err != nil {
-		return true, err
+	subjectSelectors := map[string]labels.Selector{
+		"prometheus":     labels.NewSelector(),
```
vrutkovs, line 26: CN is expected to be `prometheus-k8s`, isn't it?

machine424: `CN=system:serviceaccount:openshift-monitoring:prometheus-k8s` for certs issued to Prometheus Pods/SA

`CN=system:serviceaccount:openshift-monitoring:metrics-server` for certs issued to Metrics-server Pods/SA

machine424: (CMO will be setting those once https://github.com/openshift/cluster-monitoring-operator/pull/2329/files is merged)

pkg/cmd/controller/csr.go
vrutkovs: definitely needed, for better logging

pkg/cmd/controller/csr.go
```diff
@@ -24,23 +22,28 @@ func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedContro
...
+	}
+
+	for subject, selector := range subjectSelectors {
+		go csr.NewCSRApproverController(
+			controllerName,
+			nil,
+			kubeClient.CertificatesV1().CertificateSigningRequests(),
+			controllerCtx.KubernetesInformers.Certificates().V1().CertificateSigningRequests(),
+			csr.NewLabelFilter(selector),
+			csr.NewServiceAccountApprover(monitoringNamespace, monitoringRequesterServiceAccountName, fmt.Sprintf("CN=system:serviceaccount:%s:%s", monitoringNamespace, subject)),
```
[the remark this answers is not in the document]
vrutkovs: Worth having `csr.NewMultipleServiceAccountApprover` for that

pkg/cmd/controller/csr.go
stlaz: Yes, but your subject is wrong as used below in `csr.NewServiceAccountApprover(),`  (missing `-k8s`). You cannot do what you're doing here.

Also, I'd rather you don't set up the selectors first and then loop through the map. In case there were other bugs below, we want to make sure the selectors are always correct.

pkg/cmd/controller/csr.go
stlaz: definitely

pkg/cmd/controller/csr.go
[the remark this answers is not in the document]
stlaz: I'll have to think about whether I like the idea of a single controller per multiple CSRs with the way the controller works today.

pkg/cmd/controller/csr.go
machine424: You're right, I'll fix that (was trying to avoid duplicating code)

stlaz (changes requested):

pkg/cmd/controller/csr.go
```diff
@@ -10,37 +11,66 @@ import (
...
+	monitoringSubjectNameLabelKey         = "metrics.openshift.io/csr.subject"
 )
 
+type monitoringCSR struct {
+	subject     string
+	certSubject string
+	selector    labels.Selector
+}
+
+func newMonitoringCSR(subjectName, subjectServiceAcount string) (*monitoringCSR, error) {
```
vrutkovs, line 26:
```suggestion
func newMonitoringCSR(subjectName, subjectServiceAccount string) (*monitoringCSR, error) {
```

pkg/cmd/controller/csr.go
```diff
@@ -10,37 +11,66 @@ import (
...
+	labelsRequirement, err := labels.NewRequirement(monitoringSubjectNameLabelKey, selection.Equals, []string{subjectName})
+	if err != nil {
+		return nil, err
+	}
+	return &monitoringCSR{
+		subject:     subjectName,
+		selector:    labels.NewSelector().Add(*labelsRequirement),
+		certSubject: fmt.Sprintf("CN=system:serviceaccount:%s:%s", monitoringNamespace, subjectServiceAcount),
```
vrutkovs, line 34:
```suggestion
		certSubject: fmt.Sprintf("CN=system:serviceaccount:%s:%s", monitoringNamespace, subjectServiceAccount),
```

pkg/cmd/controller/csr.go
```diff
@@ -10,37 +11,66 @@ import (
...
+
+// RunCSRApproverController approves the CSRs for prometheus and metrics-server client certs that are used for authn against various /metrics endpoints.
 func RunCSRApproverController(ctx context.Context, controllerCtx *EnhancedControllerContext) (bool, error) {
 	kubeClient, err := controllerCtx.ClientBuilder.Client(infraClusterCSRApproverControllerServiceAccountName)
 	if err != nil {
 		return true, err
 	}
 
-	selector := labels.NewSelector()
-	labelsRequirement, err := labels.NewRequirement(monitoringLabelKey, selection.Equals, []string{monitoringLabelValue})
```
vrutkovs, line 28: Perhaps we should use `labelsRequirement, err := labels.NewRequirement(monitoringLabelKey, selection.In, []string{prometheusValue,metricsServerValue})` instead of two controllers?

vrutkovs (changes requested):

pkg/cmd/controller/csr.go
```diff
@@ -10,37 +11,66 @@ import (
...
+		return nil, err
+	}
+	return &monitoringCSR{
+		subject:     subjectName,
+		selector:    labels.NewSelector().Add(*labelsRequirement),
+		certSubject: fmt.Sprintf("CN=system:serviceaccount:%s:%s", monitoringNamespace, subjectServiceAcount),
+	}, nil
+}
+
+// RunCSRApproverController approves the CSRs for prometheus and metrics-server client certs that are used for authn against various /metrics endpoints.
```
machine424, line 38: should I mention that this runs two controllers now?

pkg/cmd/controller/csr.go
machine424: If you really mean "controllers":
We'll need two controllers because each corresponding CSRApprover can only handle one SA.
(same as https://github.com/openshift/cluster-policy-controller/pull/148#discussion_r1580831905)

If you mean "selectors":
I've already try that https://github.com/openshift/cluster-policy-controller/compare/bee77685938fb3fd9f78444cd8fbd2c282392f21..a881be203921d0d69023d97878cc91f85ccf409d and the controllers will step on each other’s toes: One will deny the requests of the other

machine424:
```
status:
  conditions:
  - lastTransitionTime: "2024-04-26T08:16:06Z"
    lastUpdateTime: "2024-04-26T08:16:06Z"
    message: expected the CSR's subject to be "CN=system:serviceaccount:openshift-monitoring:metrics-server",
      but it is "CN=system:serviceaccount:openshift-monitoring:prometheus-k8s"
    reason: CSRDenied
    status: "True"
    type: Denied
```

vrutkovs: Oh, right, serviceaccounts are different too.

I think ideally we'll want to rework this:
* `NewMultipleServiceAccountApprover` function accepting an array of serviceaccounts
* selector with `In`
* rework to use single controller

WDYT?

pkg/cmd/controller/csr.go
vrutkovs: Worth adding, yes

pkg/cmd/cluster-policy-controller/cmd.go
```diff
@@ -18,7 +19,7 @@ const (
 )
 
 func NewClusterPolicyControllerCommand(name string) *cobra.Command {
-	cmd := controllercmd.NewControllerCommandConfig("cluster-policy-controller", clusterpolicyversion.Get(), RunClusterPolicyController).
+	cmd := controllercmd.NewControllerCommandConfig("cluster-policy-controller", clusterpolicyversion.Get(), RunClusterPolicyController, clock.RealClock{}).
```
machine424, line 22: TODO: mention those changes

vrutkovs (approved):
Thanks!
/lgtm

machine424:
/retitle MON-3866: chore: csr: Add approver for monitoring CSRs issued for metrics-server

pkg/security/controller/namespace_security_allocation_controller_test.go
```diff
@@ -55,7 +57,7 @@ func TestController(t *testing.T) {
 		rangeAllocationClient: securityclient.SecurityV1(),
 		encoder:               encoder,
 	}
-	syncContext := factory.NewSyncContext(controllerName, events.NewInMemoryRecorder(controllerName))
+	syncContext := factory.NewSyncContext(controllerName, events.NewInMemoryRecorder(controllerName, clock.RealClock{}))
```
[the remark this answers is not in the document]
machine424: I didn't look deeply into it, but I've seen `clock.RealClock{}` being used in tests upstream. It's always better to avoid mocking in tests when possible, so I went with that.

Changed.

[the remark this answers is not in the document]
atiratree: Yeah, it is like that in a lot of places. Nevertheless the general guidance is to prepare for the ability to test the time better.

PR #153, opened 2024-09-12
AUTH-537: pkg/psalabelsyncer: switch to PSA version 'latest'

ibihim:
# What

Switching the label from `v1.24` to `latest`.

# Why

We are making this change for two main reasons:

1. To avoid using an outdated version label.
2. To reduce maintenance overhead.

By using the `latest` tag, we ensure that our references always point to the most recent version without requiring manual updates.


atiratree:
per https://github.com/openshift/cluster-policy-controller/pull/153#issuecomment-2345639216
/approve

PR #161, opened 2025-03-06
[WIP] CNTRLPLANE-55: Add MinimallySufficientPodSecurityStandard annotation to PSA Label Sync Controller

kramaranya:
[CNTRLPLANE-55](https://issues.redhat.com/browse/CNTRLPLANE-55)


pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
```diff
@@ -420,6 +420,7 @@ func TestEnforcingPodSecurityAdmissionLabelSynchronizationController_sync(t *tes
 		expectNSUpdate     bool
 		expectedPSaLevel   string
 		expectedPSaVersion string
+		expectedAnnotation string
```
everettraven, line 423: Is the expected annotation always going to match the expected PSa level? If so, it is probably not necessary to add an additional field in the test type here.

pkg/psalabelsyncer/podsecurity_label_sync_controller.go
```diff
@@ -372,6 +372,13 @@ func (c *PodSecurityAdmissionLabelSynchronizationController) syncNamespace(ctx c
 		}
 	}
 
+	if ns.Annotations[securityv1.MinimallySufficientPodSecurityStandard] != string(psaLevel) {
+		nsApplyConfig = nsApplyConfig.WithAnnotations(map[string]string{
+			securityv1.MinimallySufficientPodSecurityStandard: string(psaLevel),
+		})
```
everettraven, line 378: Nit: Do we need the `nsApplyConfig =` here? Everywhere else values get updated on the apply configuration it looks like its done without it.

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
kramaranya: I thought so too, and it usually matches the expected PSa level, but not always. For example, in this [namespace](https://github.com/openshift/cluster-policy-controller/blob/master/pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go#L70) the labels stay `privileged` because they're managed by someone else, but the controller computes the effective level as `restricted` and writes that into the annotation. Therefore they don't match in this [test case](https://github.com/openshift/cluster-policy-controller/pull/161/files#diff-0fb2f9db0258012fade528f2d7abeb890360c0b1b1490723e98dc2ae739bba3bR534-R535). That's why we need a separate field in the tests.

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
```diff
@@ -787,6 +802,7 @@ func TestPodSecurityAdmissionLabelSynchronizationController_sync(t *testing.T) {
 			if nsModified != nil && len(tt.expectedPSaLevel) > 0 {
 				require.Equal(t, tt.expectedPSaLevel, nsModified.Labels[psapi.WarnLevelLabel], "unexpected PSa warn level")
 				require.Equal(t, tt.expectedPSaLevel, nsModified.Labels[psapi.AuditLevelLabel], "unexpected PSa audit level")
+				require.Equal(t, tt.expectedAnnotation, nsModified.Annotations[securityv1.MinimallySufficientPodSecurityStandard], "unexpected PSa annotation value")
```
everettraven, line 805: Should this be moved to a conditional for the `expectedAnnotation` field rather than the `expectedPSaLevel`?

everettraven (approved):
changes LGTM

/lgtm

probably need @ibihim for approval

everettraven:
Nevermind, need @atiratree or @ingvagabund for approval :)

pkg/psalabelsyncer/podsecurity_label_sync_controller_test.go
ibihim: This is the exact use-case for this whole annotation :smile:

ingvagabund:
Approving based on transitively trusting @ibihim this is good to go

/approve

PR #165, opened 2025-08-11
NO-JIRA: scc allocation controller: Use server-side apply

tchap:
Move away from strategic merge patch to use SSA, which actually supersedes strategic merge patch.

This patch makes the controller not only more up-to-date, but also simpler.

On top of that, ResourceVersion is now also being passed to the server, which means conflicting updates are now prevented, as much as they were hypothetical in the past.

There is no direct reason for this refactoring, I just decided to improve this when debugging an issue for a customer.

From what I understand, the change should be compatible with the current way managed field records are stored for the namespace resource, the only difference is that when using SSA, the operation is called `Apply`, not `Update`. On top of that, the annotations are never changed anyway once set, so the namespaces updated using strategic merge patch are never going to be updated using SSA at a later time.

Feel free to reject the update, but I find the code nicer.


pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -194,40 +189,40 @@ func (c *NamespaceSCCAllocationController) allocate(ctx context.Context, syncCtx
...
+			},
+		},
 	}
-	nsCopy.Annotations[securityv1.UIDRangeAnnotation] = block.String()
-	nsCopy.Annotations[securityv1.SupplementalGroupsAnnotation] = block.String()
-	if _, ok := nsCopy.Annotations[securityv1.MCSAnnotation]; !ok {
+	// We need to make sure all annotations are always set for server-side apply, otherwise some would get removed.
+	// The following code actually makes sure the annotation is always set once it is set, which is ok regarding SSA.
+	if v, ok := ns.Annotations[securityv1.MCSAnnotation]; ok {
+		patch.Annotations[securityv1.MCSAnnotation] = v
```
ingvagabund, line 210: `patch.ObjectMetaApplyConfiguration.Annotations` to avoid any "hmm, is it really embedded?" question.

pkg/security/controller/namespace_scc_allocation_controller.go
```diff
@@ -194,40 +189,40 @@ func (c *NamespaceSCCAllocationController) allocate(ctx context.Context, syncCtx
...
+	} else {
 		if label := c.mcsAllocator(block); label != nil {
-			nsCopy.Annotations[securityv1.MCSAnnotation] = label.String()
+			patch.Annotations[securityv1.MCSAnnotation] = label.String()
```
ingvagabund, line 213: ditto

PR #167, opened 2025-08-26
OCPBUGS-57191: fix(psalabelsyncer): return an error instead of panic when converting an unknown volume

ingvagabund:
Reported through https://issues.redhat.com/browse/OCPBUGS-57191. Nevertheless, it does not seem to fix the actual underlying issue.


ingvagabund:
/label acknowledge-critical-fixes-only

A very simple fix removing the panic => improves the code stability

